PR #53, opened 2023-06-29
chore: add prettierrc and format

httpjamesm:
- Adds a `.prettierrc` config to the root directory of the project so that new contributors' formatting extensions will adhere to the existing tab spacing, quotation formality, etc.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,6 @@
+{
+  "printWidth": 80,
+  "tabWidth": 2,
+  "singleQuote": false,
+  "trailingComma": "all"
+}
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,5 +1,3 @@
 {
-    "recommendations": [
-        "ionic.ionic"
-    ]
+  "recommendations": ["ionic.ionic"]
 }
--- a/server.mjs
+++ b/server.mjs
@@ -27,7 +27,7 @@ const validLemmyServers = {};
 const badLemmyServers = {};
 
 INITIAL_VALID_LEMMY_SERVERS.forEach(
-  (server) => (validLemmyServers[server] = true)
+  (server) => (validLemmyServers[server] = true),
 );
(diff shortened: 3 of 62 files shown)


aeharding:
Hello! Thanks very much for the PR!

A couple comments:

 - wefwef is using prettier's defaults (without a `.prettierrc`). It's what [elk](https://github.com/elk-zone/elk) does too, I believe. I would like to keep using those defaults, if possible!
 - I now realize that there is not build step in the CI to check prettier, so I should probably add that 😅 (unless you want to?)

httpjamesm:
This prettier config uses all the defaults, but also adds the trailing commas requirement. My prettier defaults have wider indentations which were messing up the formatting.

If you're not up for the trailing commas, it can be removed.

aeharding:
That's strange. Prettier doesn't have a global configuration, I'm pretty sure.

> Prettier intentionally doesn’t support any kind of global configuration. This is to make sure that when a project is copied to another computer, Prettier’s behavior stays the same. Otherwise, Prettier wouldn’t be able to guarantee that everybody in a team gets the same consistent results.

https://prettier.io/docs/en/configuration.html

Maybe there is an environment issue? Or do you use editorconfig and maybe that's breaking something? Could you try an editorconfig with root = true and see if that resolves your issues?

httpjamesm:
I don't have an editor config and adding `root = true` didn't work. Not quite sure...

httpjamesm:
Ah I see. It's because I set the tab width in my VSCode prettier extension to 4 some time ago. I usually use 4 for my other projects, so maybe it's still beneficial to keep a prettier config here?

aeharding:
Yeah, that's fine. We could also add an editorconfig with indent size. As long as it doesn't change the existing code formatting.

aeharding:
I'm going to close this - feel free to open a new PR with the prettier file (if it doesn't change existing code formatting)

PR #71, opened 2023-06-29
New icon

kra-mo:
I've been really enjoying the app, so I thought I would give creating an icon for it a try!


aeharding:
This looks awesome! Some thoughts

- We need a 512x512 to make Chrome happy. Can you provide that too?
- Do you have an SVG? (It's OK if not.)
- Do you have an transparent mascot without the green bg? Then I can add lock screen images using [pwa-asset-generator](https://github.com/elegantapp/pwa-asset-generator)
- We need `maskable` variants of the 256 and 512 images. [Info](https://web.dev/maskable-icon/) - [Test site](https://maskable.app/). TLDR a variant with more padding for Android icons that have much more masking
- To be clear, are you relinquishing all rights to this logo to me?

Sorry, I know that's a lot!

kra-mo:
I messaged you on Matrix so we can discuss this.

kra-mo:
Possibly something like this if the app is to be renamed to Voyager.

logo_512

fer0n:
@aeharding are you actively looking for a different icon than the current one?

aeharding:
Yes, I thought it would be fun to have this be community driven!

I'm going to close this PR for now. Let's discuss further in #262

PR #74, opened 2023-06-29
feat: Dark mode settings

jmaruiz:
Adds two toggles to the Settings > Appearance page
1. Whether or not to obey the system value for light/dark mode
2. If user chooses not to use the system value, whether or not to always use dark mode

toggles
toggles2

Code Details:
- Changed CSS to rely on class based dark mode theming rather than media query alone
- Implemented logic to store and update toggle values from settings screen
- Implemented change detection for system dark/light theme if toggle is enabled

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -38,18 +38,19 @@ export default function App() {
   return (
     <AppContextProvider>
       <Provider store={store}>
-        <GlobalStyles />
-        <BeforeInstallPromptProvider>
-          <UpdateContextProvider>
-            <Router>
-              <IonApp>
-                <Auth>
-                  <TabbedRoutes />
-                </Auth>
-              </IonApp>
-            </Router>
-          </UpdateContextProvider>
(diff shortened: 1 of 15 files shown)


aeharding:
Hi, thanks for the PR!

Can you update this to use a radio that shows/hides depending on the "use system" slider? Like apollo. There's an Ionic component that makes this easy - IonRadio

image

jmaruiz:
Absolutely! I've never used Ionic before so thanks for the direction.

jmaruiz:
toggles3

jmaruiz:
> Hi, thanks for the PR!
> 
> Can you update this to use a radio that shows/hides depending on the "use system" slider? Like apollo. There's an Ionic component that makes this easy - IonRadio
> 
> image

Much happier with the results now, let me know what you think!

aeharding:
Looks great @jmaruiz!

Because this touches global stylesheets, I need to review in detail probably tomorrow or sat. Thanks again :)

PR #100, opened 2023-06-30
Add emergency hotfix for rejected connections to lemmy.world

aeharding:
Background:

Yesterday, reverse proxied requests on lemmy.app started being rejected from wefwef.app due to rate limiting built into lemmy on lemmy.world due to a sustantial increase in traffic over the last 48 hours.

lemmy.world will shortly enable CORS.

Note: This hotfix may only be deployed to wefwef.app.

Future versions of lemmy will enable CORS * by default like Mastodon, which will completely resolve this issue.


aeharding:
This is blocked until lemmy.world enables CORS

PR #106, opened 2023-06-30
Typo fix on new post popup

brendn:
Just a small fix - I noticed that when you make a new post the popup would say "Posted created!".  Updated it to "Post created!".


aeharding:
Thanks for the PR! :)

PR #121, opened 2023-07-01
feature: add cursor:pointer to post buttons

### LINKED ISSUE #115: Clickable areas should use "cursor:pointer" css, for better desktop browser UX
This is of course not an issue on mobile, but on desktop browsers, the text selection cursor is appearing throughout the interface, and a "pointer" (finger hand) cursor would be more appropriate.

tpkee:
Fixes #115

### DIFF
--- a/src/features/post/actions/ActionButton.tsx
+++ b/src/features/post/actions/ActionButton.tsx
@@ -1,9 +1,13 @@
 import styled from "@emotion/styled";
 
-export const ActionButton = styled.div`
+export const ActionButton = styled.button`
   display: flex;
   align-items: center;
   justify-content: center;
+
+  background: inherit;
+  font-size: inherit;
+  color: inherit;
 
   padding: 0.35rem;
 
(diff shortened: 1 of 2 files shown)


aeharding:
Hi @tpkee thanks!

This looks great, however we could change these to <IonButton> (or just try normal html button if not easily workable) for better semantic html.

Do you want to try that? If not, I'll merge this in and take a look later.

tpkee:
> Hi @tpkee thanks!
> 
> This looks great, however we could change these to (or just try normal html button if not easily workable) for better semantic html.
> 
> Do you want to try that? If not, I'll merge this in and take a look later.

Yeah sure, that's even better!

tpkee:
> Hi @tpkee thanks!
> 
> This looks great, however we could change these to (or just try normal html button if not easily workable) for better semantic html.
> 
> Do you want to try that? If not, I'll merge this in and take a look later.

Done. I also modified the MoreActions component so that its options icon is not just an icon but an actual button.

aeharding:
LGTM!

Issue #124, opened 2023-07-01
Wrong Servers on self hosted instance when installed


gianmarcotoso:
I have create a self-hosted instance with a simple docker-compose.yml file:

```
version: '3'

services:
    wefwef:
        image: ghcr.io/aeharding/wefwef:latest
        container_name: wefwef
        ports:
            - ${WEFWEF_PORT:-5314}:5314
        environment:
            - PUID=1000
            - PGID=1000
            - TZ=${TZ:-Europe/Rome}
            - CUSTOM_LEMMY_SERVERS=${WEFWEF_SERVERS:-lemmy.world,feddit.it}
        restart: unless-stopped
```

The value of `CUSTOM_LEMMY_SERVERS` is read correctly when navigating to said instance with a browser (either desktop or mobile, Firefox or Chrome), but if I install the PWA the server list changes to what I assume is the default list (lemmy.world, lemmy.ml, beehaw.org, sh.itjust.works); the website to which the PWA points to is my own instance, and not another one.

**STEPS TO REPRODUCE**
1. Create a `docker-compose.yml` file with the above contents
2. Create an `.env` file in the same directory and fill in the values (optional, not creating the file uses the default values specified in the docker-compose.yml)
3. Create an nginx reverse proxy to a domain or subdomain that points to the server you're working on, restart nginx, create the certificates with certbot
4. Start the container with `docker-compose up -d`
5. Navigate to your instance, see correct server list
6. Install the PWA, see default (?) server list

gianmarcotoso:
I think I found the problem, the value is injected through a transformer in the `server.mjs` file, but when the app is installed it's not loaded through the server, so the transformer is not there. I might try a fix or two later, the first thing that come to mind are

- Set `CUSTOM_LEMMY_SERVERS` using `define` in `vite.config.ts`
- Change `CUSTOM_LEMMY_SERVERS` to `VITE_CUSTOM_LEMMY_SERVERS` so that it's exposed to the WebApp and maybe build the production version of the app and serve that version (I haven't looked into VitePWA or ViteExpress though, maybe this already happens)

gianmarcotoso:
Reopening as the previous MR did not fix the issue (see the MR for more context)

aeharding:
@gianmarcotoso I think I got a working solution for this in #434!

PR #127, opened 2023-07-01
Fix issue #124 wrong servers when installed

gianmarcotoso:
Fixes Issue #124 

Defines the `CUSTOM_LEMMY_SERVERS` as a variable through `vite.config.ts`, allowing it to be present after PWA installation.

This requires reading the the environment via Vite's `loadEnv` function, and define the config using a function in order to be able to read the current `mode` (`development`, `production`, `test`).

The `CUSTOM_LEMMY_SERVERS` can be defined in the environment either system wide or through a `.env` file such as `.env.production.local` or `.env.development.local` (the `.local` is not strictly needed, but `.local` files are gitignored).

Sidenote: I added a `.prettierrc` because I use different settings and they were wreaking havoc on the code, this way it should be consistent.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -60,7 +60,8 @@
     {
       "files": ["src/**"],
       "globals": {
-        "APP_VERSION": true
+        "APP_VERSION": true,
+        "CUSTOM_LEMMY_SERVERS": true
       }
     }
   ]
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,5 @@
+{
+	"tabWidth": 2,
+	"useTabs": false,
+	"semi": true
+}
--- a/src/helpers/lemmy.ts
+++ b/src/helpers/lemmy.ts
@@ -1,8 +1,8 @@
 import { Comment, CommentView, Community } from "lemmy-js-client";
 
 export const LEMMY_SERVERS =
-  "CUSTOM_LEMMY_SERVERS" in window
-    ? (window.CUSTOM_LEMMY_SERVERS as string[])
+  typeof CUSTOM_LEMMY_SERVERS !== "undefined"
+    ? (CUSTOM_LEMMY_SERVERS as string).split(",")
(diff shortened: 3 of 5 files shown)


aeharding:
Looks great!

aeharding:
Hi @gianmarcotoso 

I forgot to test this in production.

Essentially, *post* build-time, you should be able to configure custom servers

But if I do:

 - `yarn build`
 - `NODE_ENV=production CUSTOM_LEMMY_SERVERS=lemmy.ml PORT=5106 node server.mjs`

The lemmy.ml server is not being picked up.

So I am reverting for now. Feel free to make another PR.

gianmarcotoso:
Hey @aeharding 

I see your point, I admit I worked under the (wrong) assumption that they should be configured at compile time. I'll look for another way, as the PWA is misbehaving anyway after installation, but yes, for now it's best to revert.

aeharding:
Thanks @gianmarcotoso!

gianmarcotoso:
@aeharding I'm thinking that there is no good solution using env in this case, as it does not carry over when installing the PWA. I was thinking about adding the ability for the user add and remove custom servers at runtime, saving them in localstorage. Would this approach be ok with you? Where should this happen, in the servers screen or in the settings screen?

Or maybe it could suffice saving the value of `CUSTOM_LEMMY_SERVERS` in localstorage, since the login info carries over after installing?

aeharding:
The custom lemmy servers should not be user configurable.

Maybe there is a something we can hook into on vite-pwa to move over custom servers?

Otherwise, we could make an express route in server.mjs to respond with servers (with aggressive 1 day caching). But that's async, so would be a bit more work.

gianmarcotoso:
Hm, I see. 

So, I've tried using localStorage to save the content of window.CUSTOM_LEMMY_SERVERS when it's available and it does work in production, so that might be a quick and viable way to solve this.

Can I ask, why shouldn't the servers be configurable (on a per-user basis) if the app lives entirely on the user's device? I'm very likely missing a piece of the puzzle here :)

PR #129, opened 2023-07-01
Build Docker images for Arm v8

Crazybus:
I'm currently building this locally on Arm v8 and the build is working perfectly! Thought I'd open this up to see if you would be interested in hosting the images to save me having to rebuild for every release.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -38,7 +38,7 @@ jobs:
         uses: docker/build-push-action@v4
         with:
           context: .
-          platforms: linux/amd64
+          platforms: linux/amd64,linux/arm/v8
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.metal.outputs.tags }}
           labels: ${{ steps.metal.outputs.labels }}


aeharding:
Would be happy to get this in! Can you check why the build is failing though?

joshdick:
Looks like an intermittent failure with the Yarn package registry, a simple rebuild may get things working again:

`232.3 error An unexpected error occurred: "https://registry.yarnpkg.com/ionicons/-/ionicons-7.1.2.tgz: ESOCKETTIMEDOUT".`

Crazybus:
> Looks like an intermittent failure with the Yarn package registry, a simple rebuild may get things working again:

Looks to be a known and not fixed issue in Yarn v1: https://github.com/yarnpkg/yarn/issues/8242. I added in the workaround of increasing the network timeout (even though the issue isn't a network timeout) to see if that helps.

joshdick:
@aeharding Looks like this should be good to merge now that the build is passing?

aeharding:
tbh I am worried about this build being flakey with the network issues. 😅

I'll merge, but I will revert if it's breaking the build

Thanks for the PR :)

Crazybus:
> I'll merge, but I will revert if it's breaking the build

Yup totally fair, I don't mind building my own at all if it becomes a pain to maintain. If you end up having to revert we could instead look at how hard it is to migrate to the v2 format where this is apparently fixed.

aeharding:
Sounds good! I also looked at changing to pnpm, which I'm totally open to switching to, but I had a problem with getting deps to work. If you want to take a shot I'm all for it!

joshdick:
🤔 Hmm, seems like I'm still not able to pull a prebuilt ARM v8 image after both this PR was merged and v0.12 was released:

```
❯ sudo docker pull ghcr.io/aeharding/wefwef:latest

latest: Pulling from aeharding/wefwef
no matching manifest for linux/arm64/v8 in the manifest list entries
```

Crazybus:
>  Hmm, seems like I'm still not able to pull a prebuilt ARM v8 image after both this PR was merged and v0.12 was released:

Just checked and it is working for me 🤔

I can also see the new image being listed here https://github.com/aeharding/wefwef/pkgs/container/wefwef.

joshdick:
Ah, looks like the difference is that my machine is looking for `linux/arm64/v8` but this PR added the platform `linux/arm/v8` (`arm` vs. `arm64`.)

The image does still run if I manually pull it this way and start a container explicitly referencing the image ID:

`sudo docker pull ghcr.io/aeharding/wefwef --platform linux/arm/v8`

But Docker warns when starting the container:

`WARNING: The requested image's platform (linux/arm/v8) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested`

Should I open a 2nd PR that also adds `linux/arm64/v8` to the `platforms` list?

mbentley:
Typically, an arm64 image would be referenced as `linux/arm64` and not `linux/arm/v8`.  See https://docs.docker.com/build/building/multi-platform/.  I am not an expert on GitHub Actions but I wouldn't think it would be any different.

aeharding:
PR welcome to address! I'd lean to renaming it instead of adding another if it solves the problem

PR #134, opened 2023-07-01
feat: Let self-hosted instances have a server whitelist

Raymonf:
I wanted to be able to only allow certain servers to be proxied. I was also having trouble setting the environment variables without the two `ENV` lines in the Dockerfile, so those lines are for that.

I'm open to further changes, if this is not satisfactory for being upstreamed. Thanks!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,5 +49,7 @@ COPY --from=builder /wefwef/dist ./dist
 EXPOSE 5314/tcp
 
 ENV PORT=5314
+ENV CUSTOM_LEMMY_SERVERS=
+ENV ALLOWED_LEMMY_SERVERS=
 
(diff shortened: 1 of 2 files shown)


aeharding:
Hi thanks for the PR. However, it is intentional to allow any server (and just the ability to customize defaults) so that people can use wefwef with their various accounts.

If you've ever used [Elk](https://github.com/elk-zone/elk) - I'm trying to take a similar approach to that.

I understand there may be concerns with reverse proxying, however future versions of Lemmy will allow CORS for direct connection. :-)

Of course, people can fork wefwef and implement the blocking, but I don't think we should merge this into the official repo at this time.

Raymonf:
Okay, fair enough :)

Thanks anyways!

PR #139, opened 2023-07-01
Setting to have comments threads collapsed by default

### LINKED ISSUE #130: Setting to have comments collapsed by default
It would be great if comment threads could be collapsed by default via a setting.

tpkee:
Closes #130 

Probably there is some (or a lot!) of work to do here. Particularly, I would check out the useEffect used in the CommentTree component. Right now I don't have a better idea, so it may be fine but I would check it out twice

### DIFF
--- a/src/features/comment/Comment.tsx
+++ b/src/features/comment/Comment.tsx
@@ -29,7 +29,7 @@ const rainbowColors = [
   "#00FFFF", // Cyan
 ];
 
-const CustomIonItem = styled(IonItem)`
+export const CustomIonItem = styled(IonItem)`
   scroll-margin-bottom: 35vh;
 
   --padding-start: 0;
@@ -38,7 +38,7 @@ const CustomIonItem = styled(IonItem)`
   --min-height: 0;
 `;
 
-const PositionedContainer = styled.div<{
+export const PositionedContainer = styled.div<{
   depth: number;
   highlighted: boolean;
 }>`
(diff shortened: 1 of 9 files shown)


aeharding:
Hi Thanks for this.

However, I think we need a different approach.

We should get only one level deep when the mode is to collapse children.

Then, the comment should use its data to determine if there are more comments to load in. It should then display how many more replies they are, like this:

image

When tapped, the comments for that node should be fetched and appended to the tree.

This would also solve problems with only being able to see 8 levels deep.

This might involve a bit of refactor, so if you would like, you can change this to just include the setting and I will merge this to a feature branch.

tpkee:
> Hi Thanks for this.
> 
> However, I think we need a different approach.
> 
> We should get only one level deep from the API when the mode is to collapse children.
> 
> Then, the comment should use its data to determine if there are more comments to load in. It should then display how many more replies they are, like this:
> 
> image
> When tapped, the comments for that node should be fetched and appended to the tree.
> 
> This would also solve problems with only being able to see 8 levels deep.
> 
> This might involve a bit of refactor, so if you would like, you can change this to just include the setting and I will merge this to a feature branch.

Oh yeah, that's much better!
I dropped the commit in which I include the functionality, leaving just the new setting option. I pushed the change so you can open the feature branch. I'll try to work on it tomorrow, but no guarantee

aeharding:
Thank you very much :)

aeharding:
I am currently working on this.

Issue #140, opened 2023-07-01
Select text button shows up when there's no text to select


aeharding:
@httpjamesm, can you take a look at this? Mainly for URL posts.

Other issues:

 1. Select text dialog scrolls up/down when trying to select text. We should disable gestures for the dialog I think
 2. Select text contents needs a bit of padding
 3. Select text should use a text area or something that behaves a bit better on desktop. Right now, it's just text in the HTML

httpjamesm:
I've addressed the first two. I don't understand the need for the third. On desktop I'm able to select just fine

aeharding:
Thanks!

When I tap once on desktop Firefox, the headers are also selected.

<img width="915" alt="image" src="https://github.com/aeharding/wefwef/assets/2166114/c3dc5df2-d117-4cf3-94cc-1a7242b3bafa">

A text area would ensure you can only drag to select the text within the text area.

httpjamesm:
Got it. I haven't tested it on Firefox.

aeharding:
Should be resolved :)

PR #142, opened 2023-07-01
Select text modal adjustments 140

httpjamesm:
Addresses #140

### DIFF
--- a/src/features/comment/Comment.tsx
+++ b/src/features/comment/Comment.tsx
@@ -254,7 +254,6 @@ export default function Comment({
                       <CommentEllipsis
                         comment={commentView}
                         rootIndex={rootIndex}
-                        selectTextModalIsOpen={selectTextModalIsOpen}
                         setSelectTextModalIsOpen={setSelectTextModalIsOpen}
                       />
                       <Ago date={comment.published} />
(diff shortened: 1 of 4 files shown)


httpjamesm:
@aeharding We should be good to go here

aeharding:
Thanks for these adjustments, I'll try to review as soon as I can

src/features/comment/Comment.tsx
```diff
@@ -197,6 +198,8 @@ export default function Comment({
   // Comment from slice might be more up to date, e.g. edits
   const comment = commentById[commentView.comment.id] ?? commentView.comment;
 
+  const [selectTextModalIsOpen, setSelectTextModalIsOpen] = useState(false);
```
aeharding, line 201: Can we avoid putting this in Comment.tsx?

Perhaps we can resolve this with #169

aeharding: (this component is already doing way too much)

httpjamesm: Hmm. I suppose we can put this out and change its value with React context. How does that sound?

aeharding:
This PR is superseded by #513 so closing. Thank you!

PR #149, opened 2023-07-02
Scroll back to top, and then back to where you were

mrmikeface:
In Apollo, a really handy feature to go back to the top on long threads is to tap the comment counter area at the top of the screen. In case of accidental tap, tapping the area again brings you back to where you were in the comments section.

In Wefwet, we don't have access to [SFSafariViewController](https://developer.apple.com/documentation/safariservices/sfsafariviewcontroller) which would enable status bar taps to scroll to the top. Instead I have used a floating action button to facilitate the same experience

### DIFF
--- a/src/features/shared/scroll/ScrollToTopFab.tsx
+++ b/src/features/shared/scroll/ScrollToTopFab.tsx
@@ -1,0 +1,28 @@
+import React from "react";
+import { IonFab, IonFabButton, IonIcon } from "@ionic/react";
+import { chevronDownOutline, chevronUpOutline } from "ionicons/icons";
+
+interface ScrollToTopFabProps {
+  onClick: () => void;
+  activated: boolean;
+}
+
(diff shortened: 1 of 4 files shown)


mrmikeface:
I noticed a couple imports and a commented out line I'll remove in a new revision. 

I forgot to add the button to the comments page but I am realizing I cannot locate which file that's in

mrmikeface:
resolves aeharding/wefwef#92

aeharding:
Hi @mrmikeface, thanks for this!

I wonder if it's possible to implement this on the status bar without the fab, now that there is a native app?

This fab may conflict with the hide posts fab, so it would be nice if it could be done without additional UI

aeharding:
I'm going to close this for now - but I'd love to see another implementation using https://capacitorjs.com/docs/apis/status-bar. That provides a callback for status bar taps (in native version) that can be used to scroll up/down as appropriate!

PR #166, opened 2023-07-02
Add appropriately scaled favicons

### LINKED ISSUE #120: Firefox Android is not using the apple-touch-icon icon.
Tested with Firefox 115.0b9

When installing wefwef as an app using Firefox it isn't taking the apple-touch-icon icon or favicon and is instead only using a w for the icon.

I've seen some old bug reports that state that a 180x180px icon is sometimes required for it to choose to use it, but I'm not sure how true that is.

### LINKED ISSUE #179: iOS splash screens
I plan on using pwa-asset-generator to do this.

I'll get started on this shortly.

nathanielcwm:
Closes #120
Closes #179?


nathanielcwm:
I've remade the SVG, and the associated PNGs, so the SVG is significantly smaller as a result.

nathanielcwm:
Looks like Firefox pulls the 180 size apple-touch-icon instead of the one defined in the manifest.

aeharding:
Hi! Thanks for this :)

If not too hard - is it possible to increase the stroke a bit of the Lemmy icon to match? Or was there a reason for the thinner stroke. Otherwise this looks fabulous!

Also planning on using this for splashes https://www.npmjs.com/package/pwa-asset-generator - basically that tool allows you to take any SVG and any background (like a linear gradient) and creates various splash PNGs for various iOS devices. Would you be interested in helping with this?

nathanielcwm:
I took the lemmy logo vector off wikimedia which took it from the lemmy Github.

What colour do you think would be best for the splash screen?

aeharding:
I'm not sure :)

I believe there can be a different color for dark vs light mode too.

Options, options... If you'd like we can just get this merged in now and discuss splash elsewhere.

nathanielcwm:
Well I would need to fix that linter error first :p

Looks like the Wikipedia licensing information is mislabelled tho - it states that the lemmy logo is creative commons but the only license I've found in the lemmy Github is AGPL, fucking copyright making things a mess again.

nathanielcwm:
Oh I thought this project was MIT licensed because I only checked the readme and not the license file lmao. Well I see db92a7f and don't have any problems now.

aeharding:
Hi @nathanielcwm 

Pushed a couple commands to automatically regenerate splashes. I don't know if that's possible with icons due to padding stuff, but LMK

A couple thoughts on icons

 - The 512.png is the correct amount of padding that should be used for the apple touch icons and non-maskable android icons
 - The 192.png is the correct amount of padding that should be used for maskable android icons *only*
 - We'll need to update manifest.json "purpose"

As you can see by the icon below on iOS, I believe it has too much padding, it appears to be using the "maskable" variety.

IMG_F5F41800BEA7-1

aeharding:
Nice work on this!!

PR #169, opened 2023-07-02
Refactor/improve login modal performance

tpkee:
This is quite a weird (and maybe pointless) PR.

The problem: some performed actions (opening the modals, for example) are quite stutter-ish on mobile (iPhone 13 Pro Max with iOS 16.5.1) which is annoying and ugly to see.

Looking through the code I saw that many modals are rendered within the post component (and subcomponents!), although they are the same for every post, so I moved (for this PR, which is a proof-of-concept more than anything) the hook to call the Login Modal up to the parent (so in the PostFeed component) and I passed down the Childs the login function. 

This is dirty-ish too in my opinion, and probably it would be best provided by a LoginModal context instead of prop drilling

[A video with before vs after ](https://imgur.com/a/2CIZDo4)

### DIFF
--- a/src/features/comment/CommentEllipsis.tsx
+++ b/src/features/comment/CommentEllipsis.tsx
@@ -24,7 +24,6 @@ import { useBuildGeneralBrowseLink } from "../../helpers/routes";
 import { useAppDispatch, useAppSelector } from "../../store";
 import { handleSelector, jwtSelector } from "../auth/authSlice";
 import { PageContext } from "../auth/PageContext";
-import Login from "../auth/Login";
 import CommentReply from "./reply/CommentReply";
 import {
   getHandle,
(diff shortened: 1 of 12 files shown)


aeharding:
This is interesting, and I definitely appreciate making animations smoother.

Perhaps we could make a dialog context? We also need to abstract the comment reply there as well, due to #182

tpkee:
> This is interesting, and I definitely appreciate making animations smoother.
> 
> Perhaps we could make a dialog context? We also need to abstract the comment reply there as well, due to #182

Yeah, that's definitely something that I would like to do. I'll try to work and finish this by tomorrow as I shouldn't have much work

aeharding:
Hi @tpkee 

I really appreciate your work on this! I am working on integrating something like this with #182 and using your headwork here

I hope that's OK. It will likely be a separate PR. I'll tag your for review, I'm trying to get this in tonight because I consider #182 very important Sorry for stepping on toes.

tpkee:
> Hi @tpkee
> 
> I really appreciate your work on this! I am working on integrating something like this with #182 and using your headwork here
> 
> I hope that's OK. It will likely be a separate PR. I'll tag your for review, I'm trying to get this in tonight because I consider #182 very important Sorry for stepping on toes.

Oh yeah sure, no problem. Glad to be of help!

aeharding:
This *should* be resolved with #225, so closing

PR #176, opened 2023-07-02
Add splash images

zackhow:
These will be shown instead of a white screen when opening the app.


aeharding:
Hi thanks! However, the icon will be changing shortly (see #166) and I plan on implementing this using [pwa-asset-generator](https://www.npmjs.com/package/pwa-asset-generator).

So I am going to close this for now.

PR #196, opened 2023-07-02
Feature/profile label setting

### LINKED ISSUE #186: Setting to change how profile tab button label is displayed
Should have an option to switch between "show instance URL", "show username" and "hide" (should static "Profile" text)

This is not prioritized, just throwing in a bug to track

tpkee:
Closes #186


rsammelson:
Could you switch the order of username and instance URL so that it is ordered from most verbose to least?

tpkee:
> Could you switch the order of username and instance URL so that it is ordered from most verbose to least?

where?

> Edit: perhaps adding another option for just username would be useful for someone who has multiple accounts across multiple instances, where two are on the same instance, so they don't have to see such a long string to disambiguate?

you mean to remove the instance's urls if the user is just on the same instance? for example:
user1:

- usernameBlabla@lemmy.world;
- usernameBlabla1@lemmy.world;
These will become simply [usernameBlaBla, usernameBlaBla1]
user2:
- username2@lemmy.world;
- username3@lemmy.ml;
These will stay username2@lemmy.world, username3@lemmy.ml 

correct?

rsammelson:
> > Could you switch the order of username and instance URL so that it is ordered from most verbose to least?
> 
> where?

In the selector in settings, but I counldn't find the code that controls that.
 
> > Edit: perhaps adding another option for just username would be useful for someone who has multiple accounts across multiple instances, where two are on the same instance, so they don't have to see such a long string to disambiguate?
> 
> you mean to remove the instance's urls if the user is just on the same instance? for example: user1:
> 
>     * [usernameBlabla@lemmy.world](mailto:usernameBlabla@lemmy.world);
> 
>     * [usernameBlabla1@lemmy.world](mailto:usernameBlabla1@lemmy.world);
>       These will become simply [usernameBlaBla, usernameBlaBla1]
>       user2:
> 
>     * [username2@lemmy.world](mailto:username2@lemmy.world);
> 
>     * [username3@lemmy.ml](mailto:username3@lemmy.ml);
>       These will stay [username2@lemmy.world](mailto:username2@lemmy.world), [username3@lemmy.ml](mailto:username3@lemmy.ml)
> 
> 
> correct?

I think so, did you mean for the domain on username2 to be different from lemmy.world?

tpkee:
> In the selector in settings, but I counldn't find the code that controls that.
> 
> > >

It is already ordered like that though? Right now it is

1. Show Instance Url
2. Show Username
3. Hide

I think so, did you mean for the domain on username2 to be different from lemmy.world?
yes, but also username3!=username2. I think it would be easier if you provided a few examples

rsammelson:
I understand now, I misread the example. That makes sense and sounds good. The main thing I want is the ability to make the name that shows up at the bottom of the screen on the profile button to be just the username.

Also, this looks kind of weird; perhaps move your setting above dark mode:

rsammelson:
> It is already ordered like that though? Right now it is
> 
>     1. Show Instance Url
> 
>     2. Show Username
> 
>     3. Hide

Somehow last time I looked (right after you opened the PR) it wasn't like that, but it looks good now.

tpkee:
> I understand now, I misread the example. That makes sense and sounds good. The main thing I want is the ability to make the name that shows up at the bottom of the screen on the profile button to be just the username.
> 
> Also, this looks kind of weird; perhaps move your setting above dark mode: 

Added the feature, please check it out

rsammelson:
I'll try running it in a few hours but this looks really nice.

rsammelson:
What changed with the light/dark settings?

Old:


New:
Screenshot 2023-07-03 at 20 47 55

Notice the sharp corners on the bottom of the top block.

rsammelson:
@aeharding Maybe it would be better to either make it a switch or make it have a page like choose navbar profile label (perhaps in preparation for more theme options). Maybe just make a "Theme" item and but the switch and the list behind it, then you could keep the current component I think.

tpkee:
> What changed with the light/dark settings?
> 
> Old: 
> 
> New: Screenshot 2023-07-03 at 20 47 55
> 
> Notice the sharp corners on the bottom of the top block.
tldr: fixed

Since I had to move the label and dark mode options under the same list, the `UserDarkMode` component was too a part of the list (since it was present in the DarkMode component). Anyhow, I moved the `UserDarkMode` component outside `DarkMode` and the list; it's not ideal imo, but I don't have a better idea

> Maybe just make a "Theme" item and but the switch and the list behind it, then you could keep the current component I think.

I guess this isn't any more needed

> Maybe it would be better to either make it a switch or make it have a page like choose navbar profile label (perhaps in preparation for more theme options).

I would leave that for another PR

rsammelson:
@tpkee can you fix the merge conflicts?

tpkee:
> @tpkee can you fix the merge conflicts?

Sorry, I've been quite busy at work lately! Right now I can't merge it, if you want you can do it yourself (obviously)

rsammelson:
@tpkee would you rather I open a PR against your branch or open my own PR on here?

rsammelson:
@tpkee I don't have permission to do that

tpkee:
Oh, open a PR against my branch I guess

tpkee:
@rsammelson merged your PR, thank you!

rsammelson:
@tpkee, apparently there are more conflicts (wow this project moves fast). I can solve them again or let you do it; I saw you are less busy but I don't have a problem doing it.

aeharding:
Thanks for keeping this up to date, I appreciate it, just looking now. I just pushed some changes.

Main things I did was move the setting to general settings to be consistent with Apollo, and I also introduced a max width on the profile tab label so it doesn't cause the tabs to get squashed in "handle" mode on mobile devices.

However, this introduces a problem:

The logic for showing instance if the instance differs with same username is an interesting idea. However, on lemmy.world vs lemmy.ml, for example, it won't show you the ending because it will be cut off.

My current pr I removed this logic and just display username. It's consistent this way and less code. However I'm looking for feedback on this, and maybe there's a way that we can convey the users' instance without attempting to show the whole handle (which doesn't really work that well on mobile, anyways)

If you feel strongly I (or you) can revert my change and use the logic you wrote. Or if you don't mind, we can just merge this as is. Either way, I think this is a great change that users will REALLY appreciate!

rsammelson:
My two cents, maybe merge this without the logic (the current version) as a "basic" implementation, and the commit with the logic will still be around if we want to spin it off into a smarter solution in a later PR as an upgrade.

tpkee:
> Thanks for keeping this up to date, I appreciate it, just looking now. I just pushed some changes.
> 
> Main things I did was move the setting to general settings to be consistent with Apollo, and I also introduced a max width on the profile tab label so it doesn't cause the tabs to get squashed in "handle" mode on mobile devices.
> 
> However, this introduces a problem:
> 
> The logic for showing instance if the instance differs with same username is an interesting idea. However, on lemmy.world vs lemmy.ml, for example, it won't show you the ending because it will be cut off.
> 
> My current pr I removed this logic and just display username. It's consistent this way and less code. However I'm looking for feedback on this, and maybe there's a way that we can convey the users' instance without attempting to show the whole handle (which doesn't really work that well on mobile, anyways)
> 
> If you feel strongly I (or you) can revert my change and use the logic you wrote. Or if you don't mind, we can just merge this as is. Either way, I think this is a great change that users will REALLY appreciate!

I think it's okay to merge this as is now, and as @rsammelson proposed make a specific PR to think about a smarter solution. This week isn't that much busy so I think I can work on it if you want, or if either you or @rsammelson or someone else want to work on it obv that's cool too!

Issue #147, opened 2023-07-01
Toast messages should not appear over tab bar


aeharding:
In Apollo, toast messages come up from *below* the tab bar and move up to appear right above it, instead of on top of it in wefwef (which is super annoying since you cannot dismiss).

This will likely require some CSS ionic hacking to change what a toast considers the bottom of the page, along with z-index hacking.

But this is absolutely a quality of life thing. :)

Anyone is welcome to take a shot at this.

Bonus points for making an ionic feature request (or finding one) to swipe away toast messages

rsammelson:
See https://github.com/ionic-team/ionic-framework/issues/21769, it looks like a PR to add swipable toasts has been languishing for a couple years.

Issue #69, opened 2023-06-29
Add ability to save media


sk4k2k:
Please add a way to save images, videos, GIFs or other media to the client's device.

aeharding:
Thanks for the issue.

Would be best as an ellipse in the media viewer, like apollo.

tomtclai:
On Apollo we can tap and hold on the image viewer to save photo

Penagwin:
@aeharding I think my PR #226 addresses the images if you wanted to merge that now.

I think saving videos might require a few extra steps, I need to test it some more but I don't think `-webkit-touch-callout:default;` alone works on the videos.

aeharding:
There's perhaps more we could do here in addition to the above, but it would best if this was actionable work in a new issue. Closing this one :)

PR #202, opened 2023-07-03
Share improvements

### LINKED ISSUE #183: Share buttons don't do anything on devices without navigator.share
e.g. desktop computer Firefox

We should probably build our own hook, useShare (or something).

It will accept exact same props as navigator.share.

If navigator.share is undefined, should copy text to clipboard (prefer URL if exists in share payload) and then present toast with useIonToast.

### LINKED ISSUE #255: Sharing posts from a feed shares the images/links, not the post itself
**Disclaimer:** I'm not sure if this is intended or not, since I haven't found any issues/discussions on this topic. The closest thing I could find was #202, which I'm not sure if it solves the matter at hand.

**Wefwef version:** 0.12.0 (installed on mobile)

### The Issue
When I use the **Share** button on a post containing an image/link from any feed - whether it's from a specific community's feed or from the Home/All/Local feeds - the shared URL corresponds to the image/link contained in the post, _not_ the post itself.
When I use the **Share** button on a post with no links or from an individual post's page (with or without images/links), the shared URL corresponds to the post itself, as I would expect.
~I made a video showing the different URLs being generated upon sharing a post with an image, followed by a post with only text - both from my Home feed:~
I *was* going to upload the video, but, for some reason, it's completely black when viewed in Github's preview...
To reproduce, you just have to go to any feed and open the sharing menu on:
- Text posts - which will share the post's URL
- Posts with images/links - which will instead share *their* URLs.

This doesn't happen on an individual post's page, where it always shares the post's URL, regardless of any existing images/links.

https://github.com/aeharding/wefwef/assets/61757466/b28a0fff-d1da-4081-b333-a776d40c7bc8

Since sharing the post itself consequentely grants access to the image/link, it seems to me that this action is more useful than sharing whatever link is contained in the post. Especially since it's possible to share the images/links by tap-holding them, whereas that doesn't seem to be an option for the post itself.

### Possible Solution
I've been looking through the code and eventually found [this line](https://github.com/aeharding/wefwef/blob/6b4758de8d1f252f383950693a3385e91309a5fa/src/features/post/shared/MoreActions.tsx#L196C22-L196C22) that I suspect could be the problem (or, at least, closely related to it). I've noticed from Chrome's network inspector that the `post` objects in the response of a simple text post don't contain a `url` property, so the code on the mentioned line is then relying on the `post.post.ap_id` property (which contains the post's URL), instead of `post.post.url` (which only exists when there is an image/link in the post). I believe sharing the `post.post.ap_id` everytime should be the correct fix, if this issue turns out to be relevant.

screenshot showing lack of URL property on a post object from a text post

 Regardless, since this is the first time I've looked at this app's code and since I've never dabbled in React - let alone React on mobile - it would be better if I let someone else tackle this.

### LINKED ISSUE #474: Add "share image" to image viewer menu
At the moment, you rely on long press on the image in order to share the image itself, not a link to the post.

For parity with Apollo, it would be great if we could get a "Share image" link in the three dots menu when the internal image viewer is open.

rsammelson:
Fixes #183 and adds modal to select what the user would like to share (link to post or link to post content). Fixes #255. Also should fix #474.

share-modal

### DIFF
--- a/src/features/gallery/GalleryPostActions.tsx
+++ b/src/features/gallery/GalleryPostActions.tsx
@@ -1,3 +1,4 @@
+import React, { useContext } from "react";
 import { IonIcon, useIonRouter } from "@ionic/react";
 import { VoteButton } from "../post/shared/VoteButton";
 import { PostView } from "lemmy-js-client";
@@ -12,9 +13,9 @@ import {
   calculateSeparateScore,
 } from "../../helpers/vote";
 import { useLocation } from "react-router";
-import React, { useContext } from "react";
 import { GalleryContext } from "./GalleryProvider";
 import { OVoteDisplayMode } from "../../services/db";
+import { Share } from "../post/shared/Share";
 
 const Container = styled.div`
   display: flex;
@@ -49,8 +50,9 @@ export default function GalleryPostActions({ post }: GalleryPostActionsProps) {
   const location = useLocation();
   const { close } = useContext(GalleryContext);
 
-  function share() {
-    navigator.share({ url: post.post.ap_id });
+  const share: React.MutableRefObject<(() => void) | null> = React.useRef(null);
+  function share_internal() {
+    share.current?.();
   }
 
   return (
@@ -70,8 +72,9 @@ export default function GalleryPostActions({ post }: GalleryPostActionsProps) {
           <Amount>{post.counts.comments}</Amount>
         </Section>
       </div>
-      <IonIcon icon={shareOutline} onClick={share} />
+      <IonIcon icon={shareOutline} onClick={share_internal} />
       <MoreActions post={post} onFeed />
+      <Share share={share} post={post} />
     </Container>
   );
 }
--- a/src/features/post/actions/PostActions.tsx
+++ b/src/features/post/actions/PostActions.tsx
@@ -5,6 +5,8 @@ import { PostView } from "lemmy-js-client";
 import { VoteButton } from "../shared/VoteButton";
 import { ActionButton } from "./ActionButton";
 import { SaveButton } from "../shared/SaveButton";
+import { Share } from "../shared/Share";
+import React from "react";
 
 const Container = styled.div`
   display: flex;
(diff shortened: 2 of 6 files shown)


rsammelson:
I'm guessing I should move the toast to the bottom. I didn't like how it covered the bottom bar, but now I realize that is a general issue that needs to be solved (#147).

rsammelson:
@aeharding This is back to being ready to merge.

rsammelson:
@aeharding, sorry for the pings, but are you ready for this PR if I fix merge conflicts, or do you want to wait?

rsammelson:
@aeharding I don't think this depends on anything else now, and I have it working with the image gallery share button too.

aeharding:
I'll take a look soon. To be honest, I'm skeptical we need to prompt the user to ask them what they would like to share 🤔 but I will try it out before any really feedback :)

aeharding:
> Oh, I like being able to share content or the post

Was this a feature in Apollo? I know there was the ability to share a picture of the post, but that would be a separate feature.

rsammelson:
I'm not sure, I never used Apollo. I just want to be able to share the post itself with people so they can interact with it (comment/vote). The app I used had a similar feature.

I assume most people aren't sharing often enough to be bothered by a two option popup when they click share, but I could add a setting to always default to sharing the media link if you want. I would really like to get this one merged so I can use it easily.

Penagwin:
Here's a screenshot of the sharing menu from Apollo - I don't remember if anything was below report.

@aeharding Do you want to combine all of the sharing and saving related meta-discussion with #69? There's a several related moving parts with them.

image

rsammelson:
In a future PR, I could add more share options to the modal created here to at least allow non-iOS users to be able to share media directly and also a share as image option. Personally I think it would be a better user experience if they only show up when clicking share, the post actions menu is already very long, and it standardizes the experience between using the share button and post actions (and I can add share as a swipe action with the same component along with customizable swipes).

Penagwin:
What I'm wondering is if we replace the "tap-hold" that iOS 16 users have with just this screen. Then it would be consistent across all platforms?

rsammelson:
Since tap and hold is a often available OS-level thing, I think leaving it is best to match other iOS apps. But I don't feel strongly about that.

rsammelson:
@aeharding I added share as a swipe option and updated from the latest main.

aeharding:
Thanks. I still need time to look at this

aeharding:
Thanks for the PR! I am closing this because it has become quite stale. Additionally, there is now an option to share the post image from the media viewer :)

PR #205, opened 2023-07-03
Pass along X-Forwarded-* headers

geneccx:
These headers help the origin web server identify the original IP that the request came from.
Otherwise, the requests will all appear to come from the wefwef server, which can be problematic for moderation and rate-limiting.

Refs:
* https://github.com/http-party/node-http-proxy#options
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For


aeharding:
Cheers!

PR #208, opened 2023-07-03
Move toast above tab bar

### LINKED ISSUE #147: Toast messages should not appear over tab bar

rsammelson:
This is currently a draft just to show a simple change that seems to at least improve the toast situation. I am also looking for feedback on if there is possibly a better way to do this.

Because of the hard coded offset some testing on various devices would be useful. This fixes #147, but not the extra feature of being able to swipe to dismiss.

### DIFF
--- a/src/GlobalStyles.tsx
+++ b/src/GlobalStyles.tsx
@@ -44,6 +44,10 @@ export default function GlobalStyles({ children }: GlobalStylesProps) {
             }
           }
 
+          ion-toast {
+            transform: translateY(-6.9ex);
+          }
+
           ${baseVariables}
 
           ${isDark ? darkVariables : lightVariables}


rsammelson:
It seems like it's not worth waiting on the swiping, so in the absence of a better approach this is probably ready to merge.

aeharding:
This has problems when added to the homescreen, or when a dialog is open. :-(

Unfortunately, I'm not sure if there is a good way to accomplish this without a ionic lib change.

rsammelson:
I will continue looking into it then.

rsammelson:
I can't find a workable solution currently.

aeharding:
Thanks for trying though @rsammelson :(

Looks like we will need to wait for an Ionic fix

PR #214, opened 2023-07-03
Clicking on feed post's image expands it

svrem:
Added the ability to expand the image of a feed post. This was a missing feature that was present in Apollo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "wefwef",
   "description": "A progressive webapp Lemmy client",
   "private": true,
-  "version": "0.11.0",
+  "version": "0.11.1",
   "type": "module",
   "scripts": {
     "dev": "node server.mjs",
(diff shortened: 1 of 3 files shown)


aeharding:
Hello. Thanks for the PR. However this was not implemented intentionally, because there are no buttons in the Preview to jump to the post comments, reply, share etc. Once those are added we can do that

svrem:
Thank you for reviewing my PR, but if I remember correctly there was no such button in Apollo either. I thought you were able to visit the detailed post by clicking on the black/white space surrounding the image. This is also the case in this PR, but please correct me if I remembered incorrectly!

svrem:
Hi, thank you for reviewing it! I indeed forgot about those so I will be looking into implementing them too.

aeharding:
Hey so an update, once #288 gets in the gallery component should be a LOT nicer :)

I added it here based on your logic but abstracted it so it's not duplicated code.

That's why I'm closing this, but I am going to add you as a co-author to my PR.

PR #123, opened 2023-07-01
Double tap search button to start search

aidankinzett:
closes #61 

Implemented using a ref

PR #220, opened 2023-07-03
feat(search-page): focus search input on search tab click

denkristoffer:
I wanted to restore a pattern from Apollo my muscle memory has been trying to use in wefwef: When the search page is shown and you click the search tab again, the input at the top is focused.

I haven't worked with Ionic before and they don't seem to provide any documentation on how to use their project with Typescript, so I did what I could but let me know if something should be changed. Specifically it seems like you shouldn't import their types, which they add to the global namespace instead, but does this work with the linter?


rsammelson:
Not to say this should be closed, but is this the same functionality as #123? If so the better implementation would need to be chosen I guess.

denkristoffer:
> Not to say this should be closed, but is this the same functionality as #123? If so the better implementation would need to be chosen I guess.

@rsammelson Oh you're right, I should've checked open PRs. The other implementation looks good to me as well.

PR #224, opened 2023-07-03
Subscribe/Unsuscribe to community in feed

fpcuriosity:
I feel we need a quicker way to subscribe to communities, especially now that people are building up their subscriptions. When scrolling through "All" or "Local" I often feel the need to quickly subscribe (or unsubscribe)


aeharding:
Hi @fpcuriosity! I've been discussing this a bit with some others - particularly on adding other options to this ellipse. It already has a ton, so we want to avoid where possible.

Currently, it's only one additional tap to subscribe, versus if it was in this menu. In fact, it might be more work if you have to scroll the ellipse menu because it has a lot of options.

I hope that makes sense on why I'm hesitant to merge this. But thanks again for the PR!

Perhaps we could add long press on community name to subscribe? That might be neat. Just thinking of other ideas.

aeharding:
Can you adjust this PR to add the option to the new long press community name action sheet?

fpcuriosity:
Yes, I'm on it. Hope to get it done by tomorrow.

Il sab 15 lug 2023, 17:11 Alexander Harding ***@***.***> ha
scritto:

aeharding:
@fpcuriosity I finished this up, thanks for getting it started!

PR #225, opened 2023-07-03
Fix comment dialogs dismissing when dirty

### LINKED ISSUE #182: Stop swiping on Mobile (or clicking outside on Desktop) from killing the comment modal/box!
I've now lost 3 longform comments I was composing because I accidentally swiped up, or scrolled up too far, and the comment modal disappears. Turns out clicking outside the comment area on Desktop also kills the comment modal 😓 When trying to comment on the same post or comment again, or trying to edit the same comment, the comment modal comes up empty.

I finally got to a point where I thought, "OK, I have to do this on desktop, this is so much time I've lost", and then I thought "OK, maybe I should just install WefWef on Desktop!". Proceeded to edit one of the comments I had previously lost all my work on, and I did something with my trackpad, and my heart sank as the bloody comment modal disappeared, taking with it an hours work of compiling a list of board games with metadata 😭

Solution suggestions:
- kill this functionality! There is a perfectly good "cancel" button right there at the top! Don't let scrolling too far on mobile, or clicking outside the comment area on desktop, make a WIP comment disappear 😔 Even flicking in the wrong direction on mobile kills the comment 😓
- warn me with a new modal, e.g. "Are you sure you want to cancel this comment?", before erasing the data. Retains the (horrible) swiping functionality while giving people a chance to not lose their work.
- auto-save WIP comments, so I can at least click "Edit" or "Comment" again and the text is still there. Certainly the least good suggestion, but would at least be something.

aeharding:
Also refactor dialogs so 'global' ones can be injected using the AppContext

Resolves #182
Related #169

@tpkee can you take a look? I'm curious if this has the performance improvements that you're expecting.

I could also break modals out into a separate ModalContext, but PageContext was already doing so little (and the `page` is for modal `presentingElement` anyways...)


aeharding:
Going to get this in, but please let me know if you have any thoughts and I can adjust :)

PR #227, opened 2023-07-04
Add button to message a user from their profile

### LINKED ISSUE #216: Send message to user from profile page
There should be a button to send a message to a user on the profile page you see when you click their name.

rsammelson:
Fixes #216

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "wefwef",
   "description": "A progressive webapp Lemmy client",
   "private": true,
-  "version": "0.16.0",
+  "version": "0.16.1",
   "type": "module",
   "scripts": {
     "dev": "node server.mjs",
--- a/src/Auth.tsx
+++ b/src/Auth.tsx
@@ -27,11 +27,9 @@ export default function Auth({ children }: AuthProps) {
   const pageVisibility = usePageVisibility();
 
   useEffect(() => {
-    if (
-      !location.pathname.startsWith("/posts") ||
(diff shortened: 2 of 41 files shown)


aeharding:
Thanks for this. However we should implement this like Apollo did - with an ellipse to open an action sheet to provide actions (block , message etc)

rsammelson:
@aeharding I think I fixed it

aeharding:
Can you use the outline variant of the mail icon, and call it "Private Message"? Otherwise looks great :)

aeharding:
Actually wait - the one other problem with this is that it changes the tab of the user. For best practice we shouldn't do that, which means the inbox stuff has to be part of a generic route (and we have to consider back button, etc)

I'm going to hold off on this release

rsammelson:
Alright, I'll take a look at fixing this in the next couple days.

rsammelson:
@aeharding how does Ionic know what tab is selected, I think I have it mostly working but when you go into it from posts it doesn't highlight any tab.

aeharding:
If the URL starts with "posts" it should be highlighted

rsammelson:
It did and there was no highlight,  but the page/highlight worked for all four other tabs, do you mind looking at some code once I get home and push it?

rsammelson:
No tab selected, the URL is `/posts/other/messages/aeharding`.

aeharding:
Should be fixed and ready to go now! :)

Issue #229, opened 2023-07-04
Support import of GDPR Reddit export


yuletide:
Love this client! I just discovered it, sadly too late to export my Apollo data now that it's closed. However I do have a complete archive export of my reddit data generated here: https://www.reddit.com/settings/data-request

Would there be a way to import those CSV files to repopulate multi-reddits and subscriptions? There is both `subscribed_subreddits.csv` and `multireddits.csv`

Barusu-rem:
> Love this client! I just discovered it, sadly too late to export my Apollo data now that it's closed. However I do have a complete archive export of my reddit data generated here: https://www.reddit.com/settings/data-request
> 
> Would there be a way to import those CSV files to repopulate multi-reddits and subscriptions? There is both `subscribed_subreddits.csv` and `multireddits.csv`
> 
> Edit: I took a first pass which is working -- if anyone knows these patterns better than I and sees issues please add comments

Faster work around set your ios date and time before July 1st (say June 29th). You should have acess to apollo export again (although you won't be able to browse reddit).

yuletide:
> > Love this client! I just discovered it, sadly too late to export my Apollo data now that it's closed. However I do have a complete archive export of my reddit data generated here: https://www.reddit.com/settings/data-request
> > Would there be a way to import those CSV files to repopulate multi-reddits and subscriptions? There is both `subscribed_subreddits.csv` and `multireddits.csv`
> > Edit: I took a first pass which is working -- if anyone knows these patterns better than I and sees issues please add comments
> 
> Faster work around set your ios date and time before July 1st (say June 29th). You should have acess to apollo export again (although you won't be able to browse reddit).

This won't work for those of us who deleted Apollo and now can't login, or for those who never used it. But yes a good option for anyone who has the app still. Otherwise this PR is helpful for everyone else.

yuletide:
Ok just tried the new multireddit flow and it is very slick

aeharding:
Closing in favor of multireddit import :)

PR #233, opened 2023-07-04
Basic settings page to import Reddit user data CSVs

yuletide:
Creating draft PR to track this work for #229 

Basic working implementation of a settings page to import subs from a reddit data export
image

Also unrelated but it would be nice to have a .prettierrc or prettier config block in the package.json for those of us who have prettier set to only format when one of those are around


aeharding:
Nice work!

yuletide:
@aeharding Thanks! Let me know if you notice anything glaring or I can take another pass before making it a full PR. I know you have lots moving rn!

aeharding:
I've updated this to import from users' multireddit link. Then they don't have to wait for reddit export request :)

Hopefully that is OK!

aeharding:
I suppose we could also have import from a reddit account data export request alongside. LMK if you feel strongly about it and we can add a generalized 'migrate' page with various migrate methods.

aeharding:
Merging for now, though!

yuletide:
@aeharding I haven't tried the multireddit approach, but if it achieves the same goal of importing all subreddits I think that is ok. We could also offer the other view as an option alongside it, for those that made their exports.

aeharding:
Yeah, I was torn with deciding that 😅

Feel free to make another issue referencing this and we can pull it in!

We'll probably want to make a generic 'import from' page with various options so we don't clutter the main settings page if we go that route, too.

yuletide:
Yeah that makes sense — import and then you can decide where from or we can
detect from the file somehow by type or headers

File input:
CSV - check for header and import as Reddit list (its just a new line
separated list anyways so could be from anything)
JSON - import as Apollo config structure

Or
Multireddit - paste your URL

yuletide:
Ok I just saw the new flow and its very slick and makes sense to just do that instead of dealing with exports unless we are importing other things also like saved multireddits

Issue #215, opened 2023-07-03
Bottom nav icons have weird artifacts (tiny spikes)


fer0n:
If you look at the bottom navigation icons closely, you can see some tiny spikes which most likely aren't there on purpose:

!["Posts" close up](https://github.com/aeharding/wefwef/assets/20093825/8c990afd-5809-45c3-8a8e-8e965a338a1c)
![Bottom Nav](https://github.com/aeharding/wefwef/assets/20093825/01be43cb-b42e-4b75-8c3f-c5818d531c22)

aeharding:
Phone model and OS version?

fer0n:
Device: iPhone 12 Pro
OS: iOS 16.6
Version: 0.11.1

aeharding:
I can replicate in https://ionic-react-conference-app.firebaseapp.com/about.

Therefore, I believe this is an Ionic or Ion Icons bug. Please make an issue with one of those projects and tag this issue so we can stay informed :)

fer0n:
@aeharding https://github.com/ionic-team/ionicons/issues/1176

aeharding:
Thanks!

If you can confirm that workaround works, feel free to make a pr

aeharding:
Reopening, I still get jagged icons

fer0n:
I'll have another look tomorrow, for it was definitely gone but maybe I was missing something. Sorry for the inconvenience

fer0n:
What device were you testing it on @aeharding?

aeharding:
I'm using an iPhone 13 mini. Should be using latest iOS :)

fer0n:
I had another look, the fix works on the iOS Simulator and on macOS in Safari. 
[Simulator Screen Shot - iPhone 13 mini](https://github.com/aeharding/wefwef/assets/20093825/c49354cc-f74a-4d28-b1b0-dfe6d8aac960).

However, no matter what I do I can't get it to work on the actual iPhone. I guess we'll have to hope that the ionicon issue will be fixed or maybe a newer iOS Version will fix it. In the mean time, if you think it's worth having that line to fix it on desktop safari you could enable it again, but otherwise I think we're out of luck, sorry.

fer0n:
Maybe there's another way to solve this, but it's probably not going to be as clean as the first approach. I'll see what solution I can come up with and then you can decide if it's worth merging

fer0n:
[...]

PR #236, opened 2023-07-04
Fix jagged ionicons in safari/webkit browsers (#215)

fer0n:
Addresses #215

### DIFF
--- a/src/theme/variables.ts
+++ b/src/theme/variables.ts
@@ -99,10 +99,6 @@ export const baseVariables = css`
     --ion-background-color: var(--ion-color-step-50, #f2f2f7);
     --ion-item-background: #fff;
   }
-
-  ion-icon {
-    -webkit-transform: translate3d(0, 0, 0);
-  }
 `;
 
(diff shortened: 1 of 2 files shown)


src/theme/variables.ts
```diff
@@ -99,6 +99,10 @@ export const baseVariables = css`
     --ion-background-color: var(--ion-color-step-50, #f2f2f7);
     --ion-item-background: #fff;
   }
+
+  ion-icon {
```
aeharding, line 103: Awesome!

Can you move this to index.css?

Sorry the naming is confusing, TBD

fer0n: I moved it and mentioned the ionicon issue in a comment. Since it's just a workaround it can be removed once/if the original issue is resolved. Thought that might help a bit.

aeharding: Thanks very much @fer0n!

aeharding:
Okay, so I probably should have tested this, but it doesn't seem to work :(

PR #240, opened 2023-07-04
Change how search bar is cleared

sasuryynala:
Add clear button to search bar so you can quickly clear text without loosing focus of the search bar. Also change cancel button to be shown when there is some text on the search bar or if the search bar is focused.


aeharding:
Perfect!

PR #244, opened 2023-07-04
skip parsing links nodes when they are duplicates

### LINKED ISSUE #73: Community links double nested <a>
Help welcome on this. But it's not a trivial bug (at least, for me). I've spent too much time in mdast/unist/unified land and a set of new eyes would be appreciated, especially if you have experience with syntax tree parsing.

https://wefwef.app/search/lemmy.world/c/asklemmy@lemmy.ml/comments/614681

Basically you get nested anchor tags because of the custom built plugin ([buildCommunityPlugin.tsx](https://github.com/aeharding/wefwef/blob/main/src/features/shared/markdown/buildCommunityPlugin.tsx)).

This code for markdown parsing of Lemmy community links is super hacky. Basically I hack the gfm autolinker to NOT parse "email-like" strings that start with !, otherwise because it always runs first, it will parse before the custom plugin can run.

Then the custom plugin runs, and it's hacky and needs refactor because it creates nested anchor elements.

see:

Screenshot 2023-06-29 at 12 32 10 PM
Screenshot 2023-06-29 at 12 32 07 PM
Screenshot 2023-06-29 at 12 32 02 PM

rideron89:
Fixes #73

Fixes an issue causing double-nesting of links. Updates the node transformer to skip transforming link nodes when the parent is of type `link` and has the same URL.

### DIFF
--- a/src/features/shared/markdown/buildCommunityPlugin.tsx
+++ b/src/features/shared/markdown/buildCommunityPlugin.tsx
@@ -35,16 +35,23 @@ export default function buildCommunityPlugin(connectedInstance: string) {
 
             const before = value.slice(lastIndex, start);
 
+            const atDomain =
+              connectedInstance === handleDomain ? "" : `@${handleDomain}`;
+            const url =
+              `https://${connectedInstance}/c/${communityHandle}${atDomain}`;
+
+            // Prevents double-nesting of links.
+            if (parent.type === "link" && parent.url === url) {
+              continue
+            }
+
             if (before) {
               newTextNodes.push({ type: "text", value: before });
             }
 
-            const atDomain =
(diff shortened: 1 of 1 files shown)


aeharding:
Thanks so much for looking at this @rideron89!

Dang, I'm still getting nested tags. :( see:

thread: /search/lemmy.world/c/asklemmy@lemmy.ml/comments/614681

image

rideron89:
Bummer! But I think I found the problem: I was comparing the URLs and they didn't always match, because we are building the URL with the connected hostname, but that doesn't match if a comment was built with a different one.

I think I'm going to have to remove that extra check for the URLs to match. Let me know if you see a problem with that.

aeharding:
Working great over here! Thanks so much for taking this over.

PR #209, opened 2023-07-03
Highlight own username

rsammelson:
Fixes #161

PR #185, opened 2023-07-02
Allow showing downvotes separately or hiding scores

rsammelson:
Fixes #170 and fixes #135. When the setting is in the default mode, it should work the same as before.

upvote<br />
downvote<br />
no votes<br />
selector

PR #246, opened 2023-07-04
Move settings to Dexie

burakcan:
This pr moves existing settings to dexie. It also adds support for any combination of user/community specific settings so we can implement some settings in a standard way. Using string keys might look weird but they're fully protected by TS so it's not possible to make a typo with those. Values are also typed per key so should be pretty safe.

Things to test:
- User installing the app for the first time.
- User updating the app from a pre-Dexie version.
- User updating the app from the last version of the DB.
- Regular flow, changing settings, seeing results of those settings

I tested these scenarios and couldn't find any issues but it's better to have another set of eyes check it.

### DIFF
--- a/src/features/settings/appearance/appearanceSlice.tsx
+++ b/src/features/settings/appearance/appearanceSlice.tsx
@@ -4,26 +4,24 @@ import {
   createSelector,
   createSlice,
 } from "@reduxjs/toolkit";
+import { merge } from "lodash";
 import { RootState } from "../../../store";
 import { MAX_DEFAULT_COMMENT_DEPTH } from "../../../helpers/lemmy";
-import { db } from "../../../services/db";
+import {
+  CommentThreadCollapse,
+  OCommentThreadCollapse,
+  OPostAppearanceType,
+  PostAppearanceType,
+  db,
+} from "../../../services/db";
+import { get, set } from "../storage";
 
-export const OPostAppearanceType = {
-  Compact: "compact",
-  Large: "large",
-} as const;
-
-export type PostAppearanceType =
-  (typeof OPostAppearanceType)[keyof typeof OPostAppearanceType];
-
-export const OCommentThreadCollapse = {
-  Always: "always",
-  Never: "never",
-  // TODO- remember per subreddit
-} as const;
-
-export type CommentThreadCollapse =
-  (typeof OCommentThreadCollapse)[keyof typeof OCommentThreadCollapse];
+export {
+  type CommentThreadCollapse,
+  type PostAppearanceType,
+  OCommentThreadCollapse,
+  OPostAppearanceType,
+} from "../../../services/db";
 
 interface AppearanceState {
   font: {
@@ -42,6 +40,17 @@ interface AppearanceState {
   };
 }
 
+const LOCALSTORAGE_KEYS = {
+  FONT: {
+    FONT_SIZE_MULTIPLIER: "appearance--font-size-multiplier",
+    USE_SYSTEM: "appearance--font-use-system",
+  },
+  DARK: {
+    USE_SYSTEM: "appearance--dark-use-system",
+    USER_MODE: "appearance--dark-user-mode",
+  },
+} as const;
+
 const initialState: AppearanceState = {
   font: {
     fontSizeMultiplier: 1,
@@ -59,6 +68,19 @@ const initialState: AppearanceState = {
   },
 };
 
+// We continue using localstorage for specific items because indexeddb is slow
+// and we don't want to wait for it to load before rendering the app and cause flickering
+const stateWithLocalstorageItems: AppearanceState = merge(initialState, {
+  font: {
+    fontSizeMultiplier: get(LOCALSTORAGE_KEYS.FONT.FONT_SIZE_MULTIPLIER),
+    useSystemFontSize: get(LOCALSTORAGE_KEYS.FONT.USE_SYSTEM),
+  },
+  dark: {
+    usingSystemDarkMode: get(LOCALSTORAGE_KEYS.DARK.USE_SYSTEM),
+    userDarkMode: get(LOCALSTORAGE_KEYS.DARK.USER_MODE),
+  },
+});
+
 export const defaultCommentDepthSelector = createSelector(
   [(state: RootState) => state.appearance.comments.collapseCommentThreads],
   (collapseCommentThreads): number => {
@@ -73,10 +95,10 @@ export const defaultCommentDepthSelector = createSelector(
 
 export const appearanceSlice = createSlice({
   name: "appearance",
-  initialState,
+  initialState: stateWithLocalstorageItems,
   extraReducers: (builder) => {
     builder.addCase(
-      fetchSettingsFromStorage.fulfilled,
+      fetchSettingsFromDatabase.fulfilled,
       (_, action: PayloadAction<AppearanceState>) => action.payload
     );
   },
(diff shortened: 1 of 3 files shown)


src/features/settings/appearance/appearanceSlice.tsx
```diff
@@ -1,25 +1,12 @@
...
-    COLLAPSE_COMMENT_THREADS: "appearance--collapse-comment-threads",
-  },
-  POSTS: {
-    TYPE: "appearance--post-type",
-  },
-  DARK: {
-    USE_SYSTEM: "appearance--dark-use-system",
-    USER_MODE: "appearance--dark-user-mode",
-  },
-} as const;
```
rsammelson, line 22: Correct me if I'm wrong, but wouldn't it be better to have one place where the keys are defined?

aeharding: Typescript is doing all the hard work to validate keys, so I think it's OK @rsammelson :)

rsammelson:
@aeharding This PR will necessitate changes to lots of other PRs (for me #185 and #209). Just for awareness, do you plan to merge this first and then update elsewhere or merge those and update here?

aeharding:
> @aeharding This PR will necessitate changes to lots of other PRs (for me #185 and #209). Just for awareness, do you plan to merge this first and then update elsewhere or merge those and update here?

Yes, this needs to get in before those. This is one of the reasons I was trying to hold off on adding additional settings for now. I probably should have communicated that more clearly in #180

Once this gets in, adding future local user & global options should be a lot more sustainable. :)

rsammelson:
Sounds good, are there other things after this they are waiting on or should I update them once this is merged?

aeharding:
That should be the only major lift in terms of dependent PRs.

src/services/db.ts
```diff
@@ -118,6 +224,99 @@ export class WefwefDB extends Dexie {
...
+    specificity?: {
+      user_handle?: string;
+      community?: string;
+    }
+  ) {
+    const { user_handle = "", community = "" } = specificity || {};
+
+    const query = this.settings
+      .where(CompoundKeys.settings.key_and_user_handle_and_community)
+      .equals([key, user_handle, community]);
```
rsammelson, line 280: Will there be settings where there is a default with the possibility of selecting something different per user? Maybe defaulting to the query with community empty followed by user and community empty would be good.

burakcan: @rsammelson this query supports all possible combinations for a setting.
- without user and without community
- with user and community
- with user but without community
- without user but with community

So if you have 2 accounts and you change such setting in one, you'll still have the default in the other one. You can also change the default for every account in theory with this but not sure how the UX would be.

rsammelson: I mean does the function automatically fall through, if so I just don't understand how the query works.

burakcan:
``` typescript
const query = this.settings
      .where(CompoundKeys.settings.key_and_user_handle_and_community)
      .equals([key, user_handle, community]);
```
the compound key in this `where` is `"[key+user_handle+community]"`. So if you
- give a "key" but leave handle and community empty; parameter to `equals` would be: `["key", "", ""]` which would match the setting which has the given "key" but doesn't have a community or handle.
- give "key" and a "handle";  parameter to `equals` would be: `["key", "handle", ""]` which would match the setting with the given "key" and "handle" but doesn't have a community. 

There can be multiple rows with the same "key" but with different handle or community values.

rsammelson: I guess I wasn't clear, I meant even if you pass in all three it might be convenient if it automatically checks for an entry that exists with only some set if there is none with all set. Does that make any sense?

burakcan: Oh I see.  Yeah it probably makes sense t o have some sort of fall back mechanism 👍

burakcan: @rsammelson added the fallback mechanism. It  tries in the following order:
- key, handle and community matches (user setting for a specific community)
- key, handle (if provided) matches but community is empty (user setting without a specific community)
- key, community (if provided) matches but handle is empty (community setting without a user)
- just the key matches (global setting)

src/services/db.ts
```diff
@@ -118,6 +224,125 @@ export class WefwefDB extends Dexie {
...
+    const { user_handle = "", community = "" } = specificity || {};
+
+    return this.transaction("r", this.settings, async () => {
+      let setting = await this.settings
+        // Everything matches
+        .where(CompoundKeys.settings.key_and_user_handle_and_community)
+        .equals([key, user_handle, community])
+        .first();
+
+      if (!setting && user_handle !== "") {
```
rsammelson, line 285: I believe this condition should be community, not user_handle (and vice versa below). That way you don't clear a string that is already empty, because that's a noop and it will be a waste of a query.

src/services/db.ts
```diff
@@ -118,6 +224,125 @@ export class WefwefDB extends Dexie {
...
+
+      if (!setting && community !== "") {
+        // Fall back to community settings if no specific setting for the user is found
+        setting = await this.settings
+          .where(CompoundKeys.settings.key_and_user_handle_and_community)
+          .equals([key, "", community])
+          .first();
+      }
+
+      if (!setting) {
```
rsammelson, line 301: This only needs to run if both user_handle and community are non-empty.

rsammelson (changes requested):
I think there's a couple small logic errors.

rsammelson:
Another option for fixing it is to  not run either of the user or community settings fallbacks unless both handles are set.

rsammelson:
@burakcan Feel free to cherry pick ebac2cf87 if you want, I just did it to make a good spot to start rebasing my other features onto.

src/services/db.ts
burakcan: I don't think I get this comment. I'm intending to do:
- If setting isn't found AND user handle is not empty,
- Try to get user setting for this key
And the same thing for the community below. Am I missing something here?

rsammelson: My point was that if community handle is empty, you already ran that query above. See the diff for ebac2cf87a.

rsammelson: Except I but the if statement on 289 around the wrong thing, it should be on 297.

rsammelson: Basically I think you should only run the inner two queries if both the user handle and community are set, otherwise the original query and the final one with both empty will cover everything.

burakcan: But when doing the first query we don't check if the community and user are empty or not. They may be non-empty when we arrive at this step. Then inside this condition we try as if community is empty to see if there is a user setting for  this key. 
I checked the diff from your branch but I still think this is correct logic. Consider following scenarios:
``` typescript
getSetting("show_nsfw", { user: "test_user", community: "test_community" }); // 1
getSetting("show_nsfw", { community: "test_community" }); // 2
getSetting("show_nsfw", { user: "test_user" }); // 3
getSetting("show_nsfw") // 4
```
1st call:
- Tries with key+handle+community, if there is a match returns it
- If there is no match AND handle is NOT empty; tries with `key+handle+[empty_community]`. This is a user setting without a specific community.
- If there is no match AND community is NOT empty; tries with `key+[empty handle]+community`. This is a community setting without a specific handle.
- If there is still no match, then tries with only the key.

2nd call:
 - Tries with key+handle+community, if there is a match returns it
 - no setting found and handle is empty, skips this step.
 - no setting found and community is NOT empty, tries with `key+[empty handle]+community`.
 - If there is still no match, then tries with only the key.

3rd call:
 - Tries with key+handle+community, if there is a match returns it
- If there is no match AND handle is NOT empty; tries with `key+handle+[empty_community]`. This is a user setting without a specific community.
 - no setting found and community is empty, skips this step.
 - If there is still no match, then tries with only the key.

4rd call:
- The first query will succeed `key+[empty handle]+[empty community]`

burakcan: Yeah in that case we do an extra unnecessary query. We can eliminate that with an additional empty check for community (or user handle in the next one). But apart from the extra query, the logic should be correct. 
Let me update that part.

burakcan: yeah @rsammelson you're right. Noticed it while updating the code. Check the new version please.

rsammelson:
I don't think that fixes the problem, we still do the extra query when there is one handle set.

burakcan:
@rsammelson the latest version should have the extra query.

rsammelson:
Almost, but now if just one is set you don't do the global setting fallback.

rsammelson (approved):

aeharding:
This looks great :)

One concern:

If your device is on dark mode, then you manually choose light mode, then reload the page, there is a 1/2 second flicker. This is quite annoying, especially on desktop.

Therefore, I am wondering if certain appearance/synchronously needed information to render the page should remain in localStorage. That includes:

 1. Font size
 2. Dark mode setting

(anything else?)

These are global, app-wide settings, and as long as we limit to synchronous settings only absolutely needed to bootstrap styles and prevent flash of content and cumulative layout shift (font size), I don't see any harm in keeping these synchronous.

burakcan:
@aeharding yeah agreed. Putting them back to localstorage

PR #261, opened 2023-07-04
Revert "Fix jagged ionicons in safari/webkit browsers (#215)"

aeharding:
Reverts aeharding/wefwef#236


aeharding:
Reverting because this doesn't seem to work

PR #264, opened 2023-07-04
Fix safari/ios 13

nickgal:
Fixes regression from https://github.com/aeharding/wefwef/pull/74

Application fails to load (blank page) with error `mediaMatch().addEventListener is not a function`.

See also: https://developer.mozilla.org/en-US/docs/Web/API/MediaQueryList#browser_compatibility

### DIFF
--- a/src/helpers/useSystemDarkMode.ts
+++ b/src/helpers/useSystemDarkMode.ts
@@ -7,20 +7,27 @@ export default function useSystemDarkMode() {
     window.matchMedia(DARK_MEDIA_SELECTOR).matches
   );
 
+  const mediaQuery = window
+    .matchMedia(DARK_MEDIA_SELECTOR);
+
   useEffect(() => {
     function handleDarkModeChange() {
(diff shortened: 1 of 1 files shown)


aeharding:
Thanks for this!

Can you please write a short inline comment on why, and/or linking to this PR? (just write "#264")

Just so it doesn't accidentally get "simplified" later 😅

nickgal:
Does that work?

aeharding:
Looks good, but prettier is complaining

src/helpers/useSystemDarkMode.ts
```diff
@@ -7,22 +7,30 @@ export default function useSystemDarkMode() {
     window.matchMedia(DARK_MEDIA_SELECTOR).matches
   );
 
+  const mediaQuery = window.matchMedia(DARK_MEDIA_SELECTOR);
```
aeharding, line 10: Why was this moved out of the effect?

Issue #267, opened 2023-07-05
Saving in compact mode opens post


rsammelson:
When you click the three dot menu and save the post in compact mode, the post is opened.

Expected behavior: it should match large mode and not open the post.

rsammelson:
Fixed in #270

PR #270, opened 2023-07-05
Fix ellipse menu taps in compact mode navigating to post (#267)

aeharding:


robertnabbe:
The `e.stopPropagation()` on `<RightDetails />` (as part of LargePost.tsx) is superfluous now, so can be removed as part of this PR:
https://github.com/aeharding/wefwef/blob/main/src/features/post/inFeed/large/LargePost.tsx#L205

aeharding:
Nice catch :)

PR #274, opened 2023-07-05
Fix #265 - Unable to search for remote users

### LINKED ISSUE #265: Cannot go to a remote user from the search page, as "@" is URI encoded as "%40" for the user search suggestion.
image

This appears to be a mistake in `/src/features/search/SearchOptions.tsx` where the variable `searchURI` is used for the user search result and not the other types. 

I have already fixed this and can submit a PR for it if that's alright with you, as per https://github.com/aeharding/wefwef/discussions/180 :)

https://github.com/bdonvr/wefwef/commit/e82253986880c51e37b1fe30dc61bc6599450a7f

bdonvr:
https://github.com/aeharding/wefwef/issues/265

looks like a simple error with the wrong variable used. :)

### DIFF
--- a/src/features/search/SearchOptions.tsx
+++ b/src/features/search/SearchOptions.tsx
@@ -33,7 +33,7 @@ export default function SearchOptions({ search }: SearchOptionsProps) {
       </InsetIonItem>
       <InsetIonItem routerLink={buildGeneralBrowseLink(`/u/${searchURI}`)}>
         <IonIcon icon={personOutline} color="primary" />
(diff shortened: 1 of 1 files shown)


aeharding:
Thanks!

mufeedali:
Would changing the label really fix the bug?

bdonvr:
> Would changing the label really fix the bug?

Nope! Guess I'm tired. Fixed the visual issue but didn't click through to test it fully... gotta change the URL to use `$search` not `$SearchURI` - Lemmy handles this fine for user profiles. Tested other special characters and it handles it gracefully. I'll do another PR

PR #277, opened 2023-07-05
Fix searching for exact communities, users

aeharding:
Resolves #276

PR #278, opened 2023-07-05
Actually fix https://github.com/aeharding/wefwef/issues/265

bdonvr:
https://github.com/aeharding/wefwef/issues/265

Need to pass $search into the URL as well - I just saw the visual bug was fixed and didn't click through to actually test this. 

IMG_1183

Not actually fixed: 
IMG_1184

Now: 
IMG_1185

Gracefully handles other special characters 

IMG_1186


aeharding:
Hi, thanks for this PR!

However I just got in #277 right before this - can you check if that resolves your issue?

Thanks!

bdonvr:
> Hi, thanks for this PR!
> 
> However I just got in #277 right before this - can you check if that resolves your issue?
> 
> Thanks!

That seems like a more comprehensive approach. I'll close this.

PR #284, opened 2023-07-05
Feature: add warning labels to compact posts

jmaruiz:
Warning labels exist on large posts, but not on compact posts, so I added them inline to hopefully not increase the size of the compact posts.


aeharding:
Thank you! I'll look at this soon

PR #292, opened 2023-07-06
Improve title search

rsammelson:
Thought it would be easier to just show what I meant. Makes prefixes more important.


aeharding (approved):
Very nice!

PR #294, opened 2023-07-06
Add support for blocking of communities 

mdecourcy:


mdecourcy:
Might also be worth it to add to the feed so users can block from all/home etc without going to the sub.

Noticed positive comments about being able to do this in other apps. Let me know if you'd like me to do that before release. Will be home from the office around 6pm

aeharding:
Added support for unblocking in the settings. I'm going to try to get this in today.

PR #315, opened 2023-07-07
Optimize Dockerfile

mbentley:
* Reduced number of layers by 3
  * Moved corepack & network-timeout to common base
  * Reduced number of RUN lines in a row
  * Remove unnecessary apk update
* Added .dockerignore to streamline build context
* Updated README to include --init (https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#handling-kernel-signals)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+*.swp
+.git/
+.github/
+.gitignore
+.prettierignore
+.vscode/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,22 @@
+# stage 0: set base image w/common customizations
 FROM docker.io/library/node:lts-alpine AS base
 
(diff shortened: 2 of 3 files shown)


aeharding:
Hi there! Thanks for the PR. However we just switched to pnpm, can you update for this?

mbentley:
Ah go figure. Will take a look. Thanks for the quick response!

mbentley:
There we go, updated for the changes.

Really like the image size difference with what I assume is the the switch over to `pnpm`:

```
REPOSITORY                  TAG          IMAGE ID       CREATED         SIZE
mbentley/wefwef             test2        0b5d71f01171   2 minutes ago   237MB
mbentley/wefwef             test1        3411ad4aca3c   3 hours ago     852MB
```

PR #322, opened 2023-07-08
Add UNPROXIED_LEMMY_SERVERS to configure bypassing the proxy

geneccx:
Since https://github.com/LemmyNet/lemmy/pull/3421 was merged, Lemmy
instances on 0.18.1 and newer allow cross-origin requests.

Expose an environment variable, UNPROXIED_LEMMY_SERVERS to support
updated servers. We use an allow-list to maintain maximum backward
compatibility while servers start to upgrade.

### DIFF
--- a/src/helpers/lemmy.ts
+++ b/src/helpers/lemmy.ts
@@ -10,6 +10,21 @@ export const LEMMY_SERVERS =
     ? (window.CUSTOM_LEMMY_SERVERS as string[])
     : ["lemmy.world", "lemmy.ml", "beehaw.org", "sh.itjust.works"];
 
+export const UNPROXIED_LEMMY_SERVERS =
+  "UNPROXIED_LEMMY_SERVERS" in window
+    ? (window.UNPROXIED_LEMMY_SERVERS as string[])
+    : [
+        "lemmy.ml",
+        "beehaw.org",
+        "sh.itjust.works",
+        "lemm.ee",
+        "feddit.de",
+        "midwest.social",
+        "lemmynsfw.com",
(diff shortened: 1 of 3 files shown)


geneccx:
Async makes this a bit of a pain... if the server is unresponsive it will hang the client so I'm revisiting the approach.

src/helpers/lemmy.ts
```diff
@@ -10,6 +10,21 @@ export const LEMMY_SERVERS =
...
+    : [
+        "lemmy.ml",
+        "beehaw.org",
+        "sh.itjust.works",
+        "lemm.ee",
+        "feddit.de",
+        "midwest.social",
+        "lemmynsfw.com",
+        "lemmy.ca",
+        "lemmy.sdf.org",
```
geneccx, line 25: I checked all these instances and they work fine without the proxy.

rsammelson: Does lemmy.world require proxy?

geneccx: Yes, it's sending double CORS headers so it breaks.

aeharding:
Let's just keep the logic in the client for now. This hack will be going away shortly anyways once an acceptable # of instances have upgraded (we're in alpha, so breaking changes are fine)

aeharding:
Sorry, you didn't have to close this, I'd like to merge this PR I just think we should keep all the logic in `src/services/lemmy.ts` for now :)

geneccx:
Ah, misunderstood what you were saying. This is now superseded by #338.

PR #327, opened 2023-07-08
Use an action sheet for post view settings instead of a full page

geneccx:
This makes it a bit easier to navigate and more closely aligns with how Apollo did it.

This should make it easier to manage when more list options are introduced, though at that point it might be worth abstracting this out into its own component.

https://github.com/aeharding/wefwef/assets/1152573/21857d63-852f-48de-bf54-2c317c1fbdb8


aeharding:
Can you use the checked css class like we have for the post sort component?

Thanks!

geneccx:
> Can you use the checked css class like we have for the post sort component?
> 
> Thanks!

Do you mean the selected role? I'm not sure what you're referring to here.

src/features/settings/appearance/PostView.tsx
```diff
@@ -1,10 +1,32 @@
...
+  OPostAppearanceType
+).map(function (postAppearanceType) {
+  return {
+    text: startCase(postAppearanceType),
+    data: postAppearanceType,
+  } as ActionSheetButton<PostAppearanceType>;
+});
 
 export default function PostView() {
+  const [open, setOpen] = useState(false);
```
aeharding, line 27: This is probably going to need to be refactored out into a separate component for each option.

Or maybe we could add an abstraction of InsetIonItem+IonActionSheet for a given option. That might be even better.

Thoughts?

We can always merge this now and deal with that later.

geneccx: I agree. If we're planning to use this for the long-term it'll be worth splitting it out. I figured that since there aren't many settings currently, a one-off was enough for now. I didn't go too far since I'm fairly new to React.

aeharding: Sounds good. We can merge this for now and abstract later

PR #326, opened 2023-07-08
Fix for issue #323: Clicking the posts button now correctly scrolls to top of the page.

svrem:
fix for issue #323

PR #329, opened 2023-07-08
Fix/scroll to top on sort change

### LINKED ISSUE #272: Page Scroll Position Not Changed After Changing Post/Comment Sort
# Issue Description:
When using the sorting feature to change the order of posts or comments on wefwef, the scroll position of the page remains the same as it was before the sort change. This behavior can lead to a poor user experience, as the user is expecting a change of context within the page and have to manually scroll back to the top.

# Environment:

- macOS chromium/ff
- Android 13: chromium/ff

# Proposed Solution:

In `wefwef/src/TabbedRoutes.ts` there is a function “scrollUpIfNeeded”. Proposing pulling this out into a helper and utilizing it as needed. Direct link to function: https://github.com/aeharding/wefwef/blob/5b254b7a956b3699de54e42dbbd36ff1362950a9/src/TabbedRoutes.tsx#L139C3-L168C1

mdecourcy:
Closes #272


src/helpers/scrollUpIfNeeded.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  if ("querySelector" in activePage) {
+    const scroll =
+      activePage?.querySelector('[data-virtuoso-scroller="true"]') ??
+      activePage
+        ?.querySelector("ion-content")
+        ?.shadowRoot?.querySelector(".inner-scroll");
+
+    if (scroll?.scrollTop) {
+      scroll.scrollTo({ top: 0, behavior: "smooth" });
```
rsammelson, line 17: Should this be scrollToIndex as below?

mdecourcy: Seems to work on my local as is, change was merged in from #326 . 

The variable scroll is also of type 'Element' which scrollToIndex does not exist on.

rsammelson (approved):

src/helpers/scrollUpIfNeeded.ts
aeharding: Yeah, `scrollToIndex` is a special Virtuoso call. It handles edge cases where Virtuoso changes height of scroll view due to virtual items of unknown height when scrolling up (and other niceties).

mdecourcy:
@aeharding I'm not home right now so can't look in to this myself, but before merging in it might be worth looking in to only scrolling to the top of the comments/bottom of the post when changing comment sort.

Could also add this in a later release.

aeharding:
> @aeharding I'm not home right now so can't look in to this myself, but before merging in it might be worth looking in to only scrolling to the top of the comments/bottom of the post when changing comment sort.
> 
> Unsure how apollo handled this, as I mainly used Infinity.
> 
> Could also add this in a later release.

That should be possible by using scrollToIndex(1) if on the post detail page!

mdecourcy:
@aeharding Thanks! Added an index var to pass-through for future use. Defaults to 0 as that will, most likely, be the primary use case.

aeharding:
Hi @mdecourcy thanks so much for this PR!

I just pushed some adjustments. Mostly because of existing buggy behavior that I've wanted to get sorted for a while now.

But also, I made it so that scroll up only occurs upon new data loaded, not before. That minimizes the "jumping". I also removed the smooth scroll for this, it seemed a little less buggy.

I'm curious what you think, so please pull and try it out!

mdecourcy:
@aeharding sorry just got back from the office. After a brief look it seems like comment sort change doesn't go up to the top on long threads. 

Also should the scroll up behavior on sort change jump up to the top, or mimic what happens on press of the "posts" tab?

aeharding:
What device are you running? It seems to be going to the top on iOS with long threads

mdecourcy:
This _was_ happening on chrome on my macbook, but after shutting down and rerunning the dev server it seems to have resolved. Looks good to me.

Going to ping you on matrix in a sec here

PR #331, opened 2023-07-08
Redirecting to posts list on submitting search bar.

svrem:
Submitting while in the search bar now results in a redirect to the posts list. This was a feature that was present in the old Apollo app that was missing.

### DIFF
--- a/src/pages/search/SearchPage.tsx
+++ b/src/pages/search/SearchPage.tsx
@@ -1,4 +1,10 @@
-import { IonHeader, IonPage, IonSearchbar, IonToolbar } from "@ionic/react";
+import {
+  IonHeader,
+  IonPage,
+  IonSearchbar,
+  IonToolbar,
+  useIonRouter,
+} from "@ionic/react";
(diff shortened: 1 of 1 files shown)


src/pages/search/SearchPage.tsx
```diff
@@ -15,23 +21,33 @@ export const focusSearchBar = () => searchBarRef.current?.setFocus();
...
-            `}
-            value={search}
-            onIonInput={(e) => setSearch(e.detail.value ?? "")}
-          />
+          <form
+            onSubmit={async (e) => {
+              e.preventDefault();
+              const el = await searchBarRef.current?.getInputElement();
+              el?.blur();
+              router.push(`/search/posts/${search}`);
```
aeharding, line 35: Please run search through encoder - `encodeURIComponent(search)` before putting on the URL

svrem: Changed it! Maybe late night coding isn't the best idea...

PR #226, opened 2023-07-03
Added support for sharing and saving images from the feed and post

Penagwin:
Addresses issue #69 for images specifically

Previously you couldn't easily press-and-hold to save an image (on iPhone) unless you selected the post, then selected the image again to be within the lightbox.

This change lets you press-and-hold to save images from the feed and post page. AFAIK this doesn't impact any of the existing gestures or clicking actions.

https://github.com/aeharding/wefwef/assets/1349689/337c851d-04fe-4671-aa4c-e062c3d98af6

PR #337, opened 2023-07-09
Allow sharing/previewing links and embeds in ios

Penagwin:
Similar to #226 it allows the web-preview/sharing for both embeds and the markdown links.
Same as the last PR - I don't think this interferes with any existing gestures.
Also helps #69

I wasn't 100% sure if the link interceptor is where you wanted this to be, or if I did it in the style you prefer.

https://github.com/aeharding/wefwef/assets/1349689/975c3492-f721-4af8-bc16-db798369020e

### DIFF
--- a/src/features/shared/markdown/LinkInterceptor.tsx
+++ b/src/features/shared/markdown/LinkInterceptor.tsx
@@ -2,13 +2,16 @@ import { LinkHTMLAttributes, MouseEvent, useCallback } from "react";
 import { useAppSelector } from "../../../store";
 import { useIonRouter } from "@ionic/react";
 import { useBuildGeneralBrowseLink } from "../../../helpers/routes";
+import styled from "@emotion/styled";
 
 const COMMUNITY_RELATIVE_URL =
   /^\/c\/([a-zA-Z0-9._%+-]+(@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,})?)\/?$/;
 
-export default function LinkInterceptor(
-  props: LinkHTMLAttributes<HTMLAnchorElement>
-) {
+const LinkInterceptor = styled(LinkInterceptorUnstyled)`
+  -webkit-touch-callout: default;
+`;
+
+function LinkInterceptorUnstyled(props: LinkHTMLAttributes<HTMLAnchorElement>) {
   const connectedInstance = useAppSelector(
     (state) => state.auth.connectedInstance
   );
@@ -63,3 +66,5 @@ function matchLemmyCommunity(
   }
   return null;
 }
+
+export default LinkInterceptor;
--- a/src/features/post/shared/Embed.tsx
+++ b/src/features/post/shared/Embed.tsx
@@ -15,6 +15,7 @@ const Container = styled.a`
 
   color: inherit;
   text-decoration: none;
+  -webkit-touch-callout: default;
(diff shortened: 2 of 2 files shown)


raziel2701:
I'm on iOS 15.7.7 and wefwef is on 0.17.0. I am still having no effect on tap and hold despite updating to this new version. When I tap and hold an image, nothing happens, I move my finger around and it takes the image around, but no menu opens up to give me the option to share the image directly or anything like what @Penagwin shared on their video.

As of right now, for me to share an image directly rather than a link, I'm taking a screenshot, cropping it and sharing that. 

https://github.com/aeharding/wefwef/assets/4183899/4a713359-c3c4-400b-8ac9-af0fc312cf40

Penagwin:
@raziel2701 When you clicked on the image that opened the "gallery" and I'll be honest I'm not sure if it's working there. The changes I made was to the image you see on the feed, and the image you see on the post, I didn't change the one in the gallery.

I should double check to make sure it's working correctly in the compact-feed.

@aeharding Would you like the ios select-hold on images to work on the gallery too?

I also noticed on my "official" pwa install that if you tap-hold really quickly then it will do the ios thing, but also open the gallery in the background. It didn't do that locally for me, is anyone able to replicate this?

Penagwin:
@raziel2701 Forgot to ask - does it work on the other images that you see on the feed and the post - not from the gallery screen?

And are you able to test my PR branch by chance and see if it's working with links for you (both embeds and blue links).

raziel2701:
@Penagwin It doesn't work anywhere I've tried inside the app. I'm not able to test your PR unfortunately.

aeharding:
> @aeharding Would you like the ios select-hold on images to work on the gallery too?

I don't see why not. I'd prefer to be consistent with any pictures in the app (that aren't small avatars)

Penagwin:
Let's do this:

I did double check and the click-hold does work on the gallery images to, so it should be consistent everywhere.

@raziel2701 I don't have a device with ios 15 - it's possible that feature isn't available on ios 15 D: 
@aeharding is this something you'd want to bypass and manually implement (with the current implementation as a stopgap), that way android users are supported to?

Let's move discussion about the images specifically to #69 

@aeharding This specific PR for the links and embeds should be good to go if you're able to test it.

src/features/shared/markdown/LinkInterceptor.tsx
```diff
@@ -2,13 +2,17 @@ import { LinkHTMLAttributes, MouseEvent, useCallback } from "react";
...
 
 const COMMUNITY_RELATIVE_URL =
   /^\/c\/([a-zA-Z0-9._%+-]+(@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,})?)\/?$/;
 
-export default function LinkInterceptor(
-  props: LinkHTMLAttributes<HTMLAnchorElement>
-) {
+const LinkInterceptor = styled(LinkInterceptorUnstyled)`
+  -webkit-touch-callout: default;
+  -webkit-tap-highlight-color: transparent;
```
aeharding, line 12: Thanks for the PR! :)

At least on my device (iPhone 13 mini) this doesn't seem to do anything

Penagwin:
@aeharding I removed the webkit-tap-highlight-color

PR #338, opened 2023-07-09
Support bypassing the proxy for some Lemmy instances

geneccx:
Since LemmyNet/lemmy#3421 was merged, Lemmy instances on 0.18.1 and newer allow cross-origin requests.

Notably missing is lemmy.world, which currently is broken as it sends double CORS headers:
```js
> fetch('https://lemmy.world/api/v3/posts/list');
```

> Access to fetch at 'https://lemmy.world/api/v3/posts/list' from origin 'https://wefwef.app' has been blocked by CORS policy: The 'Access-Control-Allow-Origin' header contains multiple values 'https://wefwef.app, *', but only one is allowed. Have the server send the header with a valid value, or, if an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.

### DIFF
--- a/src/services/lemmy.ts
+++ b/src/services/lemmy.ts
@@ -1,10 +1,22 @@
 import { LemmyHttp } from "lemmy-js-client";
 import { reduceFileSize } from "../helpers/imageCompress";
 
+const UNPROXIED_LEMMY_SERVERS = [
+  "lemmy.ml",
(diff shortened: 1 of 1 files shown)


aeharding:
Added fix for image upload. Otherwise LGTM!

PR #372, opened 2023-07-11
Fix: Voting, viewing an image or visiting a URL should set a post as read

### LINKED ISSUE #336: Feature: Voting or viewing post image from the main feed should mark post as read
Basically title, currently when in the main feed if a user upvotes/downvotes or click on a post image, the post isn't marked as read. 
Changing the behavior to mark a post as read on these interactions in combination with "hide read posts" makes it very simple to quickly go through a feed and mark all as read.

I believe Jerboa already does this for votes, but not for image click.

mdecourcy:
Resolves #336

### DIFF
--- a/src/features/gallery/GalleryPostActions.tsx
+++ b/src/features/gallery/GalleryPostActions.tsx
@@ -3,7 +3,7 @@ import { VoteButton } from "../post/shared/VoteButton";
 import { PostView } from "lemmy-js-client";
 import { chatbubbleOutline, shareOutline } from "ionicons/icons";
 import styled from "@emotion/styled";
-import { useAppSelector } from "../../store";
+import { useAppDispatch, useAppSelector } from "../../store";
 import { useBuildGeneralBrowseLink } from "../../helpers/routes";
 import { getHandle } from "../../helpers/lemmy";
 import MoreActions from "../post/shared/MoreActions";
@@ -11,6 +11,7 @@ import { calculateCurrentVotesCount } from "../../helpers/vote";
 import { useLocation } from "react-router";
 import { useContext } from "react";
 import { GalleryContext } from "./GalleryProvider";
+import { setPostRead } from "../post/postSlice";
 
 const Container = styled.div`
   display: flex;
@@ -46,6 +47,7 @@ export default function GalleryPostActions({ post }: GalleryPostActionsProps) {
   const score = calculateCurrentVotesCount(post, postVotesById);
   const location = useLocation();
   const { close } = useContext(GalleryContext);
+  const dispatch = useAppDispatch();
 
   function share() {
     navigator.share({ url: post.post.ap_id });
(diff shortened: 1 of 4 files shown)


src/features/post/postSlice.ts
```diff
@@ -194,9 +194,10 @@ export const voteOnPost =
         score: vote,
         auth: jwt,
       });
+      dispatch(setPostRead(postId));
```
aeharding, line 197: please move this into finally block

src/features/post/shared/Embed.tsx
```diff
@@ -71,14 +73,20 @@ interface EmbedProps {
 
 export default function Embed({ post, className }: EmbedProps) {
   const [error, setError] = useState(false);
+  const dispatch = useAppDispatch();
+
+  const handleLinkClick = (e: { stopPropagation: () => void }) => {
```
aeharding, line 78: `e: MouseEvent` (imported from React)

src/features/gallery/GalleryPostActions.tsx
```diff
@@ -60,6 +62,7 @@ export default function GalleryPostActions({ post }: GalleryPostActionsProps) {
       </Section>
       <div
         onClick={() => {
+          dispatch(setPostRead(post.post.id));
```
aeharding, line 65: This should not be needed. Post marked as read when you open the detail page

src/features/gallery/GalleryProvider.tsx
```diff
@@ -86,8 +89,10 @@ export default function GalleryProvider({ children }: GalleryProviderProps) {
       animationType?: PreparedPhotoSwipeOptions["showHideAnimationType"]
     ) => {
       if (lightboxRef.current) return;
+      if (!post) return;
```
aeharding, line 92: Please don't guard here, some non post components may use this. Instead, `if(post) { do the dispatching }`

aeharding (commented):
Thanks for the PR, just a bit of feedback

src/features/post/postSlice.ts
mdecourcy: Will do

src/features/gallery/GalleryPostActions.tsx
mdecourcy: My bad, didn't ctrl + s this file after removing

mdecourcy:
New commits should resolve @aeharding

src/features/post/postSlice.ts
```diff
@@ -196,8 +196,10 @@ export const voteOnPost =
       });
     } catch (error) {
       dispatch(updatePostVote({ postId, vote: oldVote }));
-
+      dispatch(setPostRead(postId));
```
aeharding, line 199: this can be removed too now that it's in the finally

aeharding (approved):
Nice job! Thanks for the quick adjustments

PR #118, opened 2023-07-01
feat: present account switcher on long pressing profile button in navbar

rahulakrishna:
https://github.com/aeharding/wefwef/assets/10240002/6e976c17-cca5-4131-b74b-17a51a83857c

- add react-hammer.js
- present account switcher `onPress`

resolves: https://github.com/aeharding/wefwef/issues/85

PR #390, opened 2023-07-12
Long press Profile tab to summon account switcher

zackhow:


aeharding:
This has similar issues to #118 - it causes weird/glitchy page animations and also breaks if you try changing instance accounts when viewing a post.

Switch accounts cross instance from every page in the app needs to be solved before this can be merged in

zackhow:
Ah, I see now. It is due to the post id's being instance specific. And when it redirects it gets a 404 since the post doesn't exist at that ID on the other instance. I'll look at this again tomorrow after work.

`path [/posts/programming.dev/c/voyagerapp@lemmy.world/comments/533712]
path [/posts/lemmy.fmhy.ml/c/voyagerapp@lemmy.world/comments/814373]
both point to ->
/posts/lemmy.world/c/voyagerapp/comments/1287799
`

zackhow:
I'm going to close this for now and if I figure it out before rahulakrishna I'll open a new PR.

Issue #325, opened 2023-07-08
[Feature Request] Notifying whether they recieved DMs or Replies properly


Rahularyan2211:
Currently, we can only see the number of inbox messages through the icon notification badge with the number. However, after going to inboxes, users don't get any info on whether they got a message, post, or comment reply.

This can be fixed by simply implementing the same technique that checks for updates used after clicking the settings button.

robertnabbe:
Picking this up.

aeharding:
Blocked until https://github.com/LemmyNet/lemmy/issues/3601

PR #407, opened 2023-07-12
feat: show unread comments/mentions/msgs in Inbox

robertnabbe:
Closes https://github.com/aeharding/voyager/issues/325.

@aeharding/@rsammelson: Unfortunately, the `/unread_count` endpoint on the lemmy back-end only includes [_Comment_ Replies, Mentions, and Private Messages](https://github.com/LemmyNet/lemmy/blob/1d38aad9d3d51ef606074d5b49a8030c49dd0e9e/crates/api/src/local_user/notifications/unread_count.rs#L23-L34)—there's no straight-forward way to get a count for Post Replies/Comments.

I went ahead and added some simple logic to determine whether a 'reply' is a POST reply or a COMMENT reply, as initially created by @aeharding in:

https://github.com/aeharding/voyager/blob/c64e84af15f563d998fbcd9bcb344d3a8222859c/src/features/inbox/InboxItem.tsx#L136-L169 

Not super elegant (ideally, this would be handled on the BE by Lemmy, but until then this will suffice). Screengrab of what this looks like:

https://github.com/aeharding/voyager/assets/13839643/381f0bea-9e15-4f02-b36c-3a6dc4caadd3

Because of the additional request that happens when the view switches to `Inbox`, there's a slight delay between showing the Bubbles—I've played with having the Bubbles be there when the view loads (use the counts from `state.inbox.counts`) and have the counts update as an alternative (code not part of PR). This looks like:

https://github.com/aeharding/voyager/assets/13839643/99f29c2e-7116-450f-9113-bbf65fcf7d36

---

I'm personally leaning toward the 'update the counts' implementation, but I'll let you (@aeharding and/or @rsammelson) make the call of what would be a better UX. Let me know.

### DIFF
--- a/src/pages/inbox/BoxesPage.tsx
+++ b/src/pages/inbox/BoxesPage.tsx
@@ -28,6 +28,13 @@ import { getInboxCounts } from "../../features/inbox/inboxSlice";
 import useClient from "../../helpers/useClient";
 import { useAppDispatch, useAppSelector } from "../../store";
 
+const UnreadBubble: FC<{ count: number }> = ({ count }) =>
+  count ? (
+    <IonBadge color="danger" slot="end">
+      {count}
+    </IonBadge>
+  ) : null;
+
 export default function BoxesPage() {
   const client = useClient();
   const jwt = useAppSelector(jwtSelector);
@@ -39,13 +46,6 @@ export default function BoxesPage() {
   } = useAppSelector((state) => state.inbox.counts);
   const [unreadPostReplies, setUnreadPostReplies] = useState(0);
   const [unreadCountsResolved, setUnreadCountsResolved] = useState(false);
-
-  const UnreadBubble: FC<{ count: number }> = ({ count }) =>
-    unreadCountsResolved && count ? (
-      <IonBadge color="danger" slot="end">
-        {count}
-      </IonBadge>
-    ) : null;
 
   const getUnreadPostReplyCount = async (): Promise<void> => {
     if (!jwt) return;
@@ -114,7 +114,7 @@ export default function BoxesPage() {
           >
             <IonIcon icon={albumsOutline} color="primary" />
             <SettingLabel>Post Replies</SettingLabel>
-            <UnreadBubble count={unreadPostReplies} />
+            {unreadCountsResolved && <UnreadBubble count={unreadPostReplies} />}
           </InsetIonItem>
           <InsetIonItem
             routerLink="/inbox/comment-replies"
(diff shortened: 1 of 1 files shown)


rsammelson:
I plan to review this code after work.

rsammelson (approved):

src/pages/inbox/BoxesPage.tsx
```diff
@@ -17,18 +18,61 @@ import {
...
   const dispatch = useAppDispatch();
+  const {
+    mentions: unreadMentions,
+    messages: unreadMessages,
+    replies: unreadReplies,
+  } = useAppSelector((state) => state.inbox.counts);
+  const [unreadPostReplies, setUnreadPostReplies] = useState(0);
+  const [unreadCountsResolved, setUnreadCountsResolved] = useState(false);
+
+  const UnreadBubble: FC<{ count: number }> = ({ count }) =>
```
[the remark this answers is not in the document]
robertnabbe: Definitely 👍

src/pages/inbox/BoxesPage.tsx
```diff
@@ -17,18 +18,61 @@ import {
...
+    replies: unreadReplies,
+  } = useAppSelector((state) => state.inbox.counts);
+  const [unreadPostReplies, setUnreadPostReplies] = useState(0);
+  const [unreadCountsResolved, setUnreadCountsResolved] = useState(false);
+
+  const getUnreadPostReplyCount = async (): Promise<void> => {
+    if (!jwt) return;
+
+    const allUnreads = await client.getReplies({
+      limit: 50,
```
aeharding, line 54: What if unreads is more than 50? I believe this will cause potentially inaccurate counts? (I've had >50 unreads a few times already).

We could potentially show >X?

Also, its unfortunate that we have to fetch 50 items just to display some counts :(

robertnabbe: Agreed on all counts (no pun intended 😭). It _is_ unfortunate we have to iterate over all unreads just to count the post replies, but I suppose it's the only option to get a valid 'post replies' count until the BE implements this count. That said, I noticed the app is fairly consistent with caching calls fwiw—this is also the reason I left the `limit` at 50, as it's unusual for the "average" user to have unreads higher than 50. There's nothing from keeping us upping this number of course; e.g. in the event a user has 500 unreads, iterating over a simple object with 500 entries without any additional processing is still peanuts (in terms of processing) for any device running this PWA.

Again, your call :)

aeharding:
Thanks again for the PR. :) I'm going to close this for now, because there's no way to do this that's not hacky.

For now, I think the best approach is for people to view "unread". The main purpose of the other buttons is to see past mentions/replies/etc *after* they're been read.

But if Lemmy adds support for distinguishing between post and comment replies in https://github.com/LemmyNet/lemmy/issues/3601 I'm happy to take another look at this :)

PR #410, opened 2023-07-12
Add experimental iOS build

aeharding:
This is experimental. It may not be merged until some other things are sorted out, and v1 is ready. I'd also like to note that *if/when this is merged* **we will continue to have a PWA** which will have its own benefits, such as more frequent updates and self-hosting.

This fixes:

- Haptics
- Tap status bar to scroll to top
- White-on-white status bar
- Scroll view sometimes locking due to [safari A2HS bug](https://bugs.webkit.org/show_bug.cgi?id=222654)

Note that this is absolutely experimental and there is more work to be done, such as

- [x] Figuring out a good way to configure dynamic scroll host depending on if installed PWA or via App Store
- [x] App icons and assets
- [ ] Bunch of signing and dealing with app store approval


mdecourcy:
@aeharding [Tauri Mobile](https://github.com/tauri-apps/tauri-mobile) may be worth looking at. Can build for both iOS + Android.

I experimented with [Tauri](https://github.com/tauri-apps/tauri) a few days ago, and was able to get a functional build out in <10 mins. (with broken functionality)

image

sosodev:
> such as more frequent updates

Is bundling the actual application data necessary? I think some webview apps still pull their content from a server to allow for the more frequent updates.

aeharding:
> > such as more frequent updates
> 
> Is bundling the actual application data necessary? I think some webview apps still pull their content from a server to allow for the more frequent updates.

Yeah, but there's compliance issues and it can be expensive. Ionic AppFlow does this but it's a minimum $50/mo

RocketRide9:
is there a prebuilt ipa package? sorry if am asking in the wrong place

PR #414, opened 2023-07-13
Fix post "more actions" menu cancel button style

rsammelson:
Cancel button no longer looks like a cancel button, it appears to be an accidental change introduced in 9514b46bb72acc65.


aeharding:
I fixed this in #413, but thanks for the PR!

PR #415, opened 2023-07-13
Fix: Link icon more noticable

### LINKED ISSUE #408: [Feature Request] Add visual hint to thumbnail type in the feed
Currently, in the compact feed there is no indication whether a thumbnail is of an image (which will open in the app) or a preview of a link which will open in a new browser window. In some reddit apps there was such an indication which made browsing a lot simpler and predictable - you knew if clicking a thumbnail would open an image, a gallery, a youtube video, or an external link.

For example, below there is no distinction between the thumbnails, but two would open an image viewer, while one will open an external browser window
no difference between thumbnails

Compared to this Boost for Reddit screen shot where you can see the small transparent triangles on every thumbnail, indicating their nature, and what would happen when I click on them 

unnamed

I'd love to see this in Voyager!

mdecourcy:
Resolves #408

Not a UX guy, so I'm very open to discussion here and won't take offense if anybody hates it.

Preview:

Before:
image

After:
image

Alternatively, but uncommitted:
image

Also uncommitted:

image

Apollo reference: (taken from above issue)
image

Summary of changes:

Set the icon color to white for better visibility, used a circular border radius, increase the opacity, added a box-shadow to create a subtle depth effect


rsammelson:
Didn't see this before I opened #416. My solution was a lot "dumber" in that I just adjust the opacity, so fixing up the general look might be the better way to go here.

I think it looks a little odd right now in dark mode on a skinny screen. Maybe too much padding?



Is this better?

mdecourcy:
@rsammelson of the two of yours, I like the skinnier better! But that's just my opinion! But I'm honestly not the guy that **anybody** goes to for UX decisions, so I think this is better decided by others.

rsammelson:
@mdecourcy the first image I took was of yours, the second is a modification of yours.

rsammelson:
@GitBarns would you like to weigh in here as well?

mdecourcy:
Also just found a compassOutline which more closely resembles apollo.

image

rsammelson:
Alright I messed with what you had some more and here's what I came up with:







I haven't committed it yet but if I should let me know.

mdecourcy:
@rsammelson they all look pretty good to me! Wonder if we should take this a step deeper, and make background/icon color dependent on  if the user has light/dark mode enabled?

rsammelson:
@mdecourcy, that's what I did. That's why it looks different between the two modes. I probably was confusing again but those pictures are all from the same code.

mdecourcy:
Oh, awesome! Yeah I'd say commit that.

rsammelson:
Alright, see https://github.com/mdecourcy/voyager/pull/3.

mdecourcy:
lgtm, merging

rsammelson:
I can't believe I forgot to format it 🤦🏼‍♂️

aeharding:
@mdecourcy I made some adjustments- switched out the icon for the "link" icon like we have in the embed component. This makes it consistent which is nice. Check out out and lmk what you think- 

image
image

mdecourcy:
Looks good to me @aeharding !

PR #416, opened 2023-07-13
Improve the link indicator for compact posts

### LINKED ISSUE #408: [Feature Request] Add visual hint to thumbnail type in the feed

rsammelson:
Fix #408

### DIFF
--- a/src/features/post/inFeed/compact/Thumbnail.tsx
+++ b/src/features/post/inFeed/compact/Thumbnail.tsx
@@ -46,15 +46,15 @@ const LinkIcon = styled(IonIcon)<{ bg: boolean }>`
   bottom: 0;
   right: 0;
   padding: 4px;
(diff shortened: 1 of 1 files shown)


rsammelson:
Closed in favor of #415

Issue #418, opened 2023-07-13
Scroll to top by clicking header


rsammelson:
This has been discussed in #92 (and other places), but I don't think it has its own issue. My understanding is that you should be able to scroll to the top by clicking on:

Communities in this image:
<img width="300" src="https://github.com/aeharding/voyager/assets/20521763/db135912-df74-41b5-9c57-2cf9003750fe">

"89 Comments" in this image:
<img width="300" alt="Screenshot 2023-07-12 at 23 40 16" src="https://github.com/aeharding/voyager/assets/20521763/73b2fd23-f6a0-46d2-8829-e9075650437e">

robertafreese:
Been reading a little about the ionic toolkit, and it looks as though there is no method provided on the ion-title component so natively it would not have any way to be a link by itself.

[Found a blog showing CSS to put a button on their page to scroll to top](https://sbsharma.com/go-to-top-button-ionic/), and I suppose it would be best as an invisible button placed over the title intentionally than the optionally visible one in this example.

rsammelson:
Thanks for the tip, see also my implementation in #419. This issue is awaiting review by @aeharding to see if it's something he wants. If it is I will implement it (more cleanly than the quick mock-up I threw together there).

aeharding:
There's a couple routes we could go:

 1. Implement tap on title. Concerns: Inconsistent UI depending on the page you're on (example: community quick switcher) and not how Apollo does it
 2. Keep current behavior. Build native iOS and Android apps that will allow tap status bar to scroll to top.

Personally I'm leaning to 2. It's consistent among all pages as well.

Thoughts on this?

robertafreese:
So, I am pretty much just just a user although I can read the code and it does look like #419 is a start to a more elegant solution than a CSS hack, of course not knowing the finer points that would be needed to move this forward.

On 1. Would this qualify as apolloparity? As a sighted user of the Community switcher, there is a down icon to give an indication of something to happen visually, so perhaps this is not unacceptably inconsistent since there is context? I have not checked yet (ionic docs or view source, but the icon probably has a screen reader label also to assist usability.

On 2. I think having native apps is clearly a smart strategy, as PWA is not well understood yet. For greater adoption, having a store presence would be optimal. Is there a toolkit in mind for this?

[...]

PR #419, opened 2023-07-13
Create demo of scrolling to top with the header

rsammelson:
Fix for #418


aeharding:
Because the native app implements this on the status bar, I am going to close this. Thanks though!

Issue #417, opened 2023-07-13
Suggestion: Remove voting from post actions menu


rsammelson:
I understand leaving reply and hide despite having swipe actions, because some people may prefer buttons. But the voting buttons, so far as I can tell, appear anywhere a post is shown. It was mentioned in #224 that the menu is getting overcrowded, which I agree with, so removing these two buttons would be an easy way to make more space.

aeharding:
In the future there will be options to remove the vote buttons from the feed like apollo.

PR #426, opened 2023-07-14
Add a "Copy Link" action in the share sheet

sharunkumar:
image


rsammelson:
This could also be moved to the share menu I'm adding in #202

sharunkumar:
> Right now the main post action sheet is almost full

if that's the case, I feel like the upvote and downvote actions can be removed from the share sheet, because they're already available as buttons 🤔

rsammelson:
See #417

PR #427, opened 2023-07-14
Blur NSFW setting

zoe-codez:
- Added `blur_nsfw` to settings database (default: on)
- Added toggle to settings page
- Created new `isNsfwBlurred` method (related to `isNsfw`)
- Updated wiring

## Settings UI update

Changed the case of an unrelated setting to make the UI more consistent


rsammelson (approved):

zoe-codez:
Good to go here. Let me know if you need any tweaks

aeharding:
Thanks! The only thing that stands out is blurring should be per user, not a global setting.

zoe-codez:
Is there another setting doing that already? Not super familiar with react, so mostly doing the copy/paste thing from other settings 😬

aeharding:
Favorites are stored like that currently. Also update to the latest main branch - there's some updates that makes that easier!

If you'd like help let me know and I will take a look. I'd like to get this in tomorrow (Saturday US)

zoe-codez:
Updated to store as user settings, seems like that part works as expected. I am observing a settings workflow issue / race condition using only user settings

The state object provided here contains the default / default on state when the browser page is initially loaded: https://github.com/aeharding/voyager/pull/427/files#diff-b9b34233b0450c7703b4926892096079610bee82258cf8710d3b6bcbef66b23aR198

It first loads from the user settings database, then it wipes that out with global settings, which interact with the same object. I added back the global setting for now to act as a "last known state" to work around that in a hacky way. Acts like a user setting though, getting tracked & swapped out with the user switcher.

aeharding:
Hi @zoe-codez I took a look and made some changes. The race condition should no longer be happening. I also changed it from a checkbox to a multi-select which cleans up the length of the label.

Can you take a look when you have a moment and let me know what you think?

Thanks!

aeharding:
(also you may want to clear your indexeddb because I made a schema change)

zoe-codez:
LGTM!

PR #428, opened 2023-07-14
Fix/Scroll to top after top > time selection

mdecourcy:
Post sort scroll change is working when sort change isn't in the nested top menu.

This PR fixes top > time sort change scroll

Before:

https://youtube.com/shorts/b-EXfg5Q7Ko

after:

https://youtu.be/_idc8nE4Igs


aeharding:
Excellent, thank you

PR #431, opened 2023-07-14
Create a StoreProvider for central coordination of side effects between slices

burakcan:
https://github.com/aeharding/voyager/pull/79#discussion_r1257393336

This can be used for fetching user specific settings from the database on user change, too. Which should enable the setting for unblurring the nsfw media.


src/features/community/communitySlice.ts
```diff
@@ -97,11 +97,15 @@ export const getFavoriteCommunities =
...
-
-    dispatch(setFavorites(communities));
+    try {
+      const communities = await db.getSetting("favorite_communities", {
+        user_handle: userHandle,
+      });
+
+      dispatch(setFavorites(communities));
+    } catch (e) {
+      // swallow the error, probably "Setting not found"
```
aeharding, line 107: btw this PR looks awesome! :) Thank you!

For this error - is it an instance of something? Can we do `if (!(e instanceof SomeError)) throw e;` so we don't hide genuine errors?

aeharding: Or perhaps a guard would look nicer

```js
} catch (error) {
  if (error instanceof SomeError) return;

  throw error;
}
```

burakcan: yeah that makes sense but currently it's a generic promise rejection error. I think that could be the topic of another pr :)

I just added this here because it was logging unhandled errors to the console.

aeharding:
I made just a couple adjustments to clear out favorites when logging out, and also when changing users to a user without favorites.

This looks awesome, thanks again

PR #436, opened 2023-07-14
Settings for compact posts

elpendor:
I doubt this is a priority but I've added two settings for compact posts:

- Thumbnail position (either left or right)
- Voting buttons visibility (toggles visibility for upvote/downvote buttons)


rsammelson:
@elpendor thank you for implementing this, I for one would really like these options. Can you fix the linting issues and undo the lock file changes so the diff reflects what actually changed?

elpendor:
> @elpendor thank you for implementing this, I for one would really like these options. Can you fix the linting issues and undo the lock file changes so the diff reflects what actually changed?

Yeah, sorry, I used npm instead of pnpm and that generated unnecesary changes to the lock files.

Also, the linting issues seems to be related to the rules working differently on windows. 

I've uploaded a new commit with the fixes and I'll create a new PR in a few minutes.

PR #438, opened 2023-07-15
Replace current community page on search

zackhow:


aeharding:
Hi @zackhow thanks for the PR. This would be the preferred approach, however it breaks the browser back button.

If you have time, could you create an upstream bug report in the Ionic project with a reproduction? If so I will make sure the right people see it :)

Thanks again. Replace is absolutely what we should be doing here, but keeping a non-broken back button is really important too.

zackhow:
I didn't notice that before. Just to confirm you mean it breaks the 'back animation'?

aeharding:
In my testing, it breaks swipe to go back completely. The gesture is just not recognized :(

zackhow:
Strange, that is not my experience. On both my mac and iphone back worked. https://imgur.com/a/2ZPBE4V

aeharding:
Ah, I see. Yes, it's just the animation that's not working.

I wonder if there's a way to get that working...?

zackhow:
Turns out the animation builder was overriding the default animations.

aeharding:
Nice, and the back button works from the community page. But now I'm seeing another problem that if you quick switch communities and then go to comments and try to swipe back, the swipe doesn't work

aeharding:
Hi @zackhow! So I looked into this a bit further and I believe the root of the issue is https://github.com/ionic-team/ionic-framework/issues/24260. Essentially, Ionic React does not respect setting the route direction ("none") when using "replace".

I have a workaround (see pushed commit) but it unfortunately does not fix the broken swipe back animation. Oh well. It does replace the current page, though!

zackhow:
@aeharding awesome! Sorry been busy these past couple weeks, it's great to get this in.

PR #439, opened 2023-07-15
Add community blocking from feed by long press

aeharding:


aeharding:
It's not in the ellipse menu, but the advantage of this approach is:

 1. You can block community anywhere you see its label, and 
 2. It doesn't add another item to the already too long post ellipse menu

PR #447, opened 2023-07-15
Change Report terminology from Custom Response to Report

eat-your-glory:
Changed the terminology of Reports from "Custom Response" to "Custom Report", as “Custom Reponse” just doesn’t sound right to me, and “Custom Report” just makes more sense.


aeharding:
Thanks for the PR. Custom Response is the Apollo terminology; I think it's ok as is :)

PR #470, opened 2023-07-16
Add setting to show instance URLs for remote users

### LINKED ISSUE #200: Option to display instance with username
Users should have an option to display other users’ instance along side their screen names.

This could be simply adding the @instance tag or via a flair kind of tag from Apollo, or even color coding/icons.

Text examples
- username@lemmy.world
- username@lemmy.ml

Flair examples
- username `lemmy.world`
- username `lemmy.ml`

rsammelson:
Broken off from #209. Fixes #200.

### DIFF
--- a/src/features/labels/links/PersonLink.tsx
+++ b/src/features/labels/links/PersonLink.tsx
@@ -5,6 +5,8 @@ import { useBuildGeneralBrowseLink } from "../../../helpers/routes";
 import { Person } from "lemmy-js-client";
 import Handle from "../Handle";
 import { StyledLink } from "./shared";
+import { useAppSelector } from "../../../store";
+import { OInstanceUrlDisplayMode } from "../../../services/db";
 
 const Prefix = styled.span`
   font-weight: normal;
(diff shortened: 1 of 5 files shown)


src/features/labels/links/PersonLink.tsx
```diff
@@ -32,6 +34,11 @@ export default function PersonLink({
 
   let color: string | undefined;
 
+  const force_instance_url =
```
aeharding, line 37: Underscores only for database properties - this can just be camel cased

src/features/settings/settingsSlice.tsx
```diff
@@ -153,50 +161,48 @@ export const appearanceSlice = createSlice({
...
       state.appearance.font.useSystemFontSize = action.payload;
-
       set(LOCALSTORAGE_KEYS.FONT.USE_SYSTEM, action.payload);
     },
+    setUserInstanceUrlDisplay(
+      state,
+      action: PayloadAction<InstanceUrlDisplayMode>
+    ) {
+      state.appearance.general.userInstanceUrlDisplay = action.payload;
+      db.setSetting("user_instance_url_display", action.payload);
```
aeharding, line 175: how about `show_user_instance`? `user_instance_url_display` is not immediately clear to me

aeharding (commented):
Couple last things. Thanks for separating this out

src/features/settings/settingsSlice.tsx
rsammelson: I named it that because my plan was to add the ability to turn them off everywhere in the future, so it will be a selector for when the user instance URL should be displayed (always, on posts [the default], never). I can still rename it if you prefer.

aeharding: It's fine, the only reason I wanted to double check is since it's in the db it'll be hard to rename in the future :)

PR #471, opened 2023-07-17
Hide read posts

aeharding:
This is a proof of concept.

TODO:

 1. ~~Scrolling user to a sensible place after hiding read posts. (Should keep track of post at top of screen that's not hidden, and scroll the user to that post)~~ Apollo just scrolls to top, seems fine.
 2. Figure out types...
 3. Cleanup UI


src/features/feed/HideReadPosts.tsx
```diff
@@ -0,0 +1,27 @@
...
+  return (
+    <IonIcon
+      icon={eyeOff}
+      onClick={async () => {
+        const postIds: number[] = itemsRef.current.map((item) => item.post.id);
+
+        const toHide = postIds.filter((id) => postReadById[id]);
+
+        for (const postId of toHide) {
+          dispatch(hidePost(postId));
```
aeharding, line 22: @burakcan currently I do a new DB insert for every hidden post. I assume it's possible to bulk add with Dexie, but if you have any pointers that would be appreciated :) it's probably necessary with the mark read on scroll feature I'm adding to the next release

aeharding: (not a huge priority though)

burakcan: @aeharding there is a bulkadd method for tables in dexie. I guess we can debounce the add calls for a couple of frames, keep the items and then bulk add them in a single transaction. I'll look into it when I have time. Did we notice any performance issues with this?

aeharding: Actually, in terms of performance it seems totally fine in a loop :) So no need to do this, unless you want to

PR #477, opened 2023-07-17
Stop Remove Account icon to be cropped

### LINKED ISSUE #403: Remove Account icon is being cropped at the top and bottom
image

As shown in the picture the top and bottom of the remove icon are slightly cut off.

Using:
Voyager 0.20.0
iPhone 13
iOS Developer Beta 3
Safari

RoToRa:
Fixes #403
The icon is now the standard size for Ionic for icon-only buttons. It's minutely smaller than before at the normal font size (23.4 x 23.4 vs 24 x 24).


aeharding:
Great, thanks!

PR #478, opened 2023-07-17
feat: added action sheet and direct link to profile in chat page

sharunkumar:
image

### DIFF
--- a/src/features/feed/ConversationsMoreActions.tsx
+++ b/src/features/feed/ConversationsMoreActions.tsx
@@ -1,9 +1,4 @@
-import {
-  IonButton,
-  IonIcon,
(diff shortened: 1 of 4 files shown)


aeharding:
Hi thanks!

Can we put this in a menu like we do in other places? See apollo example below. You can copy `SpecialFeedMoreActions.tsx` if you merge main.

IMG_3193

sharunkumar:
> Can we put this in a menu like we do in other places?

Hey, so currently, the top corner button is for "mark all as read". so if I do this, should that also be moved to the menu?

sharunkumar:
@aeharding its done now :+1: 

image

aeharding:
Looks great!

PR #485, opened 2023-07-18
Block select text for deleted comments

zoe-codez:
This PR removes the "Select Text" option for deleted / removed comments

### DIFF
--- a/src/features/comment/CommentEllipsis.tsx
+++ b/src/features/comment/CommentEllipsis.tsx
@@ -1,3 +1,4 @@
+import styled from "@emotion/styled";
 import {
   IonActionSheet,
   IonIcon,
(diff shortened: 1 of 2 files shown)


src/features/comment/CommentEllipsis.tsx
```diff
@@ -87,6 +88,23 @@ export default function MoreActions({ comment, rootIndex }: MoreActionsProps) {
...
+  let selectTextOption = undefined;
+  if (commentExists) {
+    selectTextOption = {
+      text: "Select Text",
+      role: "select-text",
+      icon: textOutline,
+    };
+  } else if (isMyComment || localMod) {
+    selectTextOption = {
+      text: "Select Deleted Text",
```
aeharding, line 103: For consistency, let's just not show "select text" if the comment isn't actually shown, and ignore moderators and "is my comment" for now. I know we'll want to revisit this with the mod work.

zoe-codez: Sounds good!

Issue #395, opened 2023-07-12
Themes


kerosyn:
Not a high priority, but it would be nice to see theme like in Apollo. I miss my Dracula theme.

PR #492, opened 2023-07-19
Add support for themes, rename existing dark theme to `Black`, and add a dark grey theme

mike-gregory:
This PR replaces the current existing theme implementation (System light/dark, light or dark) which uses boolean values with a more versatile theme implementation. It still supports the existing system functionality which will automatically choose between light/black theme depending upon your settings.

I've renamed the existing dark theme to Black, as it more closely resembles the `Black (AMOLED)` themes of other Android apps. In its place, I've added my own dark theme which uses a dark grey colour.

I've also simplified the theme selection UI to appear like this:

Theme

I'm happy to update wording to `System (Light/Black)` or `Automatic` if it causes confusion.

### ThemeProvider
In GlobalStyles.tsx was this code:

```tsx
return (
    <ThemeProvider theme={{ dark: isDark }}>
      <Global
  // ...
```

As dark mode is no longer a boolean, I've had to remove this. It doesn't appear to make any difference. I've tried toggling this value to true/false, but it didn't seem to have an effect. I've removed it for now, but happy to re-add if there's something I'm missing.

### Automatic light/dark
I've been unable to test if the "system" theme is working as intended as it never worked for me before. All of my devices are on dark mode, and switching to light doesn't cause Voyager to change **but** it doesn't change on live for me either. If someone could check for me that'd be greatly appreciated.

### Existing settings
This implementation doesn't migrate existing settings, and will default users to System.

## Gallery
| Communities                                                                                               | Posts                                                                                               | Comments                                                                                               | Settings                                                                                               | Appearance                                                                                               |
|-----------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------|
| Communities | Posts | Comments | Settings | Appearance |

Mostly implements #395, but does not implement Darcula.


aeharding:
Hi! I appreciate the PR, however there are some architectural decisions made in this PR that haven't been discussed prior.

For example, in Apollo you could have a "green" theme that's also with pure black mode. This is because pure black mode and theme are separate toggles.

We should discuss this more in #395 before proceeding :)

aeharding:
@mike-gregory I took a stab at better flexibility of theming - check out #548. It provides the flexibility I discussed above to have pure black mode for each given theme.

I like your grey theme here, however! Would you like to make a PR against my branch for that?

I left that part unimplemented - see `buildDarkVariables` of `variables.ts` of my branch. There's an unused `pureBlack` flag we can use to toggle between pure black styles.

Let me know if this is something you'd be willing to help out on! :)

aeharding:
Okay so an update, I added a non-true-black dark mode, but implemented it close to Apollo's look (a more blue-ish dark mode).

Check it out and let me know what you think. I'm also totally up for adding a more grey dark mode theme as an option, too!

aeharding:
Closing in favor of #548, but I'm happy to accept PRs for additional themes :)

PR #497, opened 2023-07-20
Add the ability to configure swipe gestures

### LINKED ISSUE #64: Configurable gestures
Apollo had configurable gestures and  I think it's a pretty useful thing to be able to customize. Screenshot from Apollo for reference: 

image

rsammelson:
Fixes #64


rsammelson:
@aeharding can you take a look at this and let me know your initial thoughts?

aeharding:
Just a cursory glance at the code, looks good from a high level. I won't be able to review until this weekend though. Thanks!

rsammelson:
Thanks!

aeharding:
Okay, so I just had some time to check this out. A couple thoughts on the UI:

 1. Let's remove the "none" option from each and replace with a slider to disable left and right sliding options (see second picture) (This will require DB changes)
 2. Change the terminology for each option to be consistent with Apollo (and sort order of the options)
 3. Add an icon to each option in the popup (see third picture)
 4. Ideally we would have a swipe icons, but I realize this requires expertise with SVGs so it can be a future thing.

IMG_3204
IMG_3205
IMG_3206

rsammelson:
>     1. Let's remove the "none" option from each and replace with a slider to disable left and right sliding options (see second picture) (This will require DB changes)

With the none option you can disable only short or long swipes, allowing you to have a single action for each direction at either the short or long swipe trigger point. This would be useful for people who want a couple of swipe actions (perhaps upvote and downvote) but want to make it harder to accidentally trigger them (by only using long swipes). While it takes a little more effort to disable swipes (I can add a disable all option if you want), it allows for a lot more configurability.
 
>     2. Change the terminology for each option to be consistent with Apollo (and sort order of the options)
>
>     3. Add an icon to each option in the popup (see third picture)

I will fix these soon.

aeharding:
> With the none option you can disable only short or long swipes, allowing you to have a single action for each direction at either the short or long swipe trigger point. This would be useful for people who want a couple of swipe actions (perhaps upvote and downvote) but want to make it harder to accidentally trigger them (by only using long swipes). While it takes a little more effort to disable swipes (I can add a disable all option if you want), it allows for a lot more configurability.

That's what the "long swipe trigger point" option does (see second screen shot)

rsammelson:
How does that option work? I don't understand which swipes it disables.

aeharding:
It doesn't disable any option, it just requires you to swipe further before changing to the second swipe option

rsammelson:
My point was that by changing individual things to None, you can set a single action for either direction, and you can choose whether you want it at the short swipe or long swipe distance. I for one would prefer not having a short swipe on comments because I accidentally trigger it sometimes, especially on the left when trying to go back.

aeharding:
I understand, however I'd rather add an option to disable short swipe left universally. Otherwise you could have inconsistent behavior on inbox items vs posts vs comments.

rsammelson:
Thoughts on this look?

aeharding:
Looks great!

rsammelson:
I added options to turn off left and right completely, as well as these icons to the lists:


<br />

<br />

rsammelson:
Future work is allowing users to disable the back gesture on iOS. Note, I did leave none as an option so far, because I would really like to have that choice; given that it's in a pretty deep menu I imagine it won't bother anyone who doesn't choose it.

src/features/shared/sliding/BaseSlidingVote.tsx
```diff
@@ -82,36 +124,162 @@ export default function BaseSlidingVote({
...
+  const all_actions: {
+    [id in SwipeAction]: SlidingItemAction | undefined;
+  } = useMemo(() => {
+    return {
+      none: undefined,
+      upvote: {
+        icon: arrowUpSharp,
         trigger: () => {
-          onVote(currentVote === 1 ? 0 : 1);
+          vote(currentVote === 1 ? 0 : 1);
```
aeharding, line 235: onVote

src/features/shared/sliding/BaseSlidingVote.tsx
```diff
@@ -82,36 +124,162 @@ export default function BaseSlidingVote({
...
+    hideAction,
+    collapseAction,
+    markUnreadAction,
+    vote,
+  ]);
+
+  const startActions: ActionList = useMemo(
+    () =>
+      !disableLeftSwipes
+        ? [all_actions[actions["start"]], all_actions[actions["far_start"]]]
```
aeharding, line 271: camelCase

src/features/shared/sliding/SlidingItem.tsx
```diff
@@ -45,19 +47,54 @@ const OptionContainer = styled.div<{ active: boolean }>`
...
+}>`
+  ${({ icon, slash, bgColor }) =>
+    slash &&
+    css`
+      &::after {
+        content: "";
+        position: absolute;
+        height: ${custom_slash_lengths[icon] ?? 30}px;
+        width: 3px;
+        background: var(--ion-color-${bgColor}-contrast);
```
aeharding, line 68: Let's always do white like before. (the bookmark should be solid white)

src/features/settings/gestures/SwipeSettings.tsx
```diff
@@ -0,0 +1,232 @@
...
+  const disableLeftSwipes = useAppSelector(
+    (state) => state.gestures.swipe.disableLeftSwipes
+  );
+  const disableRightSwipes = useAppSelector(
+    (state) => state.gestures.swipe.disableRightSwipes
+  );
+
+  return (
+    <>
+      <SwipeList
```
aeharding, line 63: Can you remove the chevrons on each swipe setting?

src/features/settings/gestures/gesturesSlice.tsx
```diff
@@ -0,0 +1,154 @@
+import { PayloadAction, createAsyncThunk, createSlice } from "@reduxjs/toolkit";
+import {
+  SwipeAction,
+  SwipeActions,
+  db,
+  default_swipe_actions_post,
```
aeharding, line 6: camelCase

src/features/shared/sliding/BaseSlidingVote.tsx
```diff
@@ -82,36 +124,162 @@ export default function BaseSlidingVote({
...
+      : undefined;
+  }, [collapsed, collapseRootComment]);
+
+  const isRead = useMemo(() => {
+    return readByInboxItemId[getInboxItemId(item as InboxItemView)] ?? false;
+  }, [item, readByInboxItemId]);
+
+  const markUnread = useCallback(async () => {
+    try {
+      await dispatch(markRead(item as InboxItemView, !isRead));
```
aeharding, line 206: Refactor to not use `as`

src/features/shared/sliding/BaseSlidingVote.tsx
aeharding, line 1: This component is doing a *lot* - and it renders for every comment and post.

I am wondering if there's a way to refactor so we only instantiate what's needed for the set custom gestures

src/services/db.ts
```diff
@@ -59,14 +59,14 @@ export type InstanceUrlDisplayMode =
 
 export const OVoteDisplayMode = {
   /**
-   * Show total score (upvotes + downvotes)
+   * Show upvotes and downvotes separately
    */
-  Total: "total",
```
aeharding, line 64: Was this a bad merge?

src/features/shared/sliding/BaseSlidingVote.tsx
```diff
@@ -1,63 +1,87 @@
-import { css } from "@emotion/react";
-import styled from "@emotion/styled";
-import { IonIcon, useIonToast } from "@ionic/react";
-import { arrowDownSharp, arrowUpSharp } from "ionicons/icons";
+import { useIonToast } from "@ionic/react";
+import {
+  arrowDownSharp,
+  arrowUndo,
+  arrowUpSharp,
+  bookmarkOutline,
```
aeharding, line 6: Solid bookmark icon

src/features/settings/gestures/swipe_long.svg
aeharding, line 1: I really like these icons!

src/features/settings/gestures/gesturesSlice.tsx
```diff
@@ -0,0 +1,154 @@
...
+  default_swipe_actions_comment,
+  default_swipe_actions_inbox,
+} from "../../../services/db";
+
+interface GestureState {
+  swipe: {
```
aeharding, line 12: Probably doesn't need to be nested in `swipe`?

src/features/inbox/inboxSlice.ts
```diff
@@ -218,7 +218,11 @@ export function getInboxItemId(item: InboxItemView): string {
     return `dm_${item.private_message.id}`;
   }
 
-  return `mention_${item.person_mention.id}`;
+  if ("person_mention" in item) {
+    return `mention_${item.person_mention.id}`;
+  }
+
+  return "";
```
aeharding, line 225: What's this for?

src/pages/settings/SettingsPage.tsx
```diff
@@ -107,6 +108,14 @@ export default function SettingsPage() {
             </IconBg>
             <SettingLabel>Appearance</SettingLabel>
           </InsetIonItem>
+
+          <InsetIonItem routerLink="/settings/gestures">
+            <IconBg color="color(display-p3 0.95 0.65 0)">
+              <IonIcon icon={returnUpForwardOutline} />
+            </IconBg>
+            <SettingLabel>Gestures</SettingLabel>
```
aeharding, line 116: Can you move this below "Filters and Blocks" to match Apollo?

aeharding:
I'm experiencing some strange behavior:

 1. Open home feed
 2. Go to gestures page
 3. Turn on "disable swipe gestures" both left and right, then turn both off
 4. Go back to feed, try to swipe item
 5. See the swipe item (upvote, reply etc) stays transparent and doesn't fully enable. You also can't swipe to the second item
 6. If you let go and swipe again, the next one should work

Can you replicate? It's not a huge bug or anything just strange behabior

src/services/db.ts
```diff
@@ -77,6 +77,62 @@ export const OVoteDisplayMode = {
...
+} as const;
+
+export type SwipeAction =
+  (typeof OSwipeActionAll)[keyof typeof OSwipeActionAll];
+
+export type SwipeDirection = "far_start" | "start" | "end" | "far_end";
+export type SwipeActions = { [id in SwipeDirection]: SwipeAction };
+
+export const default_swipe_actions_post: SwipeActions = {
+  far_start: OSwipeActionPost.Downvote,
```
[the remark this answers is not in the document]
rsammelson: I write too much C++, I'll fix the snake case.

aeharding:
I'm wondering if we should change hide post to save as the default for right swipe item. That would match Apollo and also would avoid confusing new users when they accidentally swipe away a post and lose it. What do you think?

src/features/inbox/inboxSlice.ts
rsammelson: No idea, maybe I merged something wrong

src/features/shared/sliding/BaseSlidingVote.tsx
rsammelson: How should I do that?

src/features/shared/sliding/SlidingItem.tsx
rsammelson: The bookmark is black because the green is light enough to make black the better contrast; with white it looks kind of weird. I can make it white but it will reduce contrast in dark mode to a bad level (1.97:1 vs black is 10.65:1).

src/features/settings/gestures/swipe_long.svg
rsammelson: Thanks

src/features/settings/gestures/gesturesSlice.tsx
rsammelson: I was going to do that so other gesture stuff would not be nested (e.g. disableBackGesture, which I am working on adding).

src/services/db.ts
rsammelson: No, I realized the sorting was strange, not its least verbose -> most verbose when you look top to bottom.

src/features/settings/gestures/SwipeSettings.tsx
rsammelson: What do you mean?

rsammelson:
There's still more to be done, but I wanted to push most of what I have.

src/features/shared/sliding/SlidingItem.tsx
aeharding: That's OK, make it white, I am going to tweak the green color in a separate PR

src/features/shared/sliding/SlidingItem.tsx
```diff
@@ -45,19 +47,54 @@ const OptionContainer = styled.div<{ active: boolean }>`
...
    * If `string`, it's passed to IonIcon as an icon value
    */
-  render: (() => React.ReactNode) | string;
+  icon: string;
   trigger: () => void;
   bgColor: string;
+  slash?: boolean;
 };
 
+export type ActionList = [
```
aeharding, line 89: Can we add some docs regarding when the first and second items represents (near and far swipe respectively)

src/features/settings/gestures/SwipeSettings.tsx
aeharding: These guys, we can just remove them for all settings that open the action sheet
IMG_7838E59B0D07-1

src/features/shared/sliding/BaseSlidingVote.tsx
aeharding: Something like this would work, then you can remove the changes to inboxSlice:

```diff
--- a/src/features/shared/sliding/BaseSlidingVote.tsx
+++ b/src/features/shared/sliding/BaseSlidingVote.tsx
@@ -30,7 +30,6 @@ import { PageContext } from "../../auth/PageContext";
 import { SwipeAction, SwipeActions } from "../../../services/db";
 import useCollapseRootComment from "../../comment/useCollapseRootComment";
 import { getInboxItemId, markRead } from "../../inbox/inboxSlice";
-import { InboxItemView } from "../../inbox/InboxItem";
 import { CommentsContext } from "../../comment/CommentsContext";
 import styled from "@emotion/styled";

@@ -38,10 +37,25 @@ const StyledItemContainer = styled.div`
   --ion-item-border-color: transparent;
 `;

+type SlideableItem =
+  | CommentView
+  | PostView
+  | PersonMentionView
+  | CommentReplyView;
+
+function isInboxItem(
+  item: SlideableItem
+): item is PersonMentionView | CommentReplyView {
+  if ("person_mention" in item) return true;
+  if ("comment_reply" in item) return true;
+
+  return false;
+}
+
 interface BaseSlidingVoteProps {
   children: React.ReactNode;
   className?: string;
-  item: CommentView | PostView | PersonMentionView | CommentReplyView;
+  item: SlideableItem;
   rootIndex?: number;
   collapsed?: boolean;
   actions: SwipeActions;
@@ -198,12 +212,14 @@ function BaseSlidingVoteInternal({
   }, [collapsed, collapseRootComment]);

   const isRead = useMemo(() => {
-    return readByInboxItemId[getInboxItemId(item as InboxItemView)] ?? false;
+    return isInboxItem(item) ? readByInboxItemId[getInboxItemId(item)] : false;
   }, [item, readByInboxItemId]);

   const markUnread = useCallback(async () => {
+    if (!isInboxItem(item)) return;
+
     try {
-      await dispatch(markRead(item as InboxItemView, !isRead));
+      await dispatch(markRead(item, !isRead));
     } catch (error) {
       present({
         message: "Failed to mark item as unread",
```

src/services/db.ts
aeharding: I'm not sure I understand, to me it makes sense to have default be at the top in general.

src/services/db.ts
```diff
@@ -77,6 +77,41 @@ export const OVoteDisplayMode = {
...
+  ...OSwipeActionPost,
+  ...OSwipeActionComment,
+  ...OSwipeActionInbox,
+} as const;
+
+export type SwipeAction =
+  (typeof OSwipeActionAll)[keyof typeof OSwipeActionAll];
+
+export type SwipeDirection = "farStart" | "start" | "end" | "farEnd";
+export type SwipeActions = { [id in SwipeDirection]: SwipeAction };
```
aeharding, line 113: Can we change this to record notation? `Record<SwipeDirection, SwipeAction>`

There's probably an eslint rule for this somewhere I could add, too...

src/features/shared/sliding/BaseSlidingVote.tsx
```diff
@@ -82,36 +124,162 @@ export default function BaseSlidingVote({
...
+    return {
+      icon: mailUnread,
+      trigger: markUnread,
+      bgColor: "primary",
+      slash: !isRead,
+    };
+  }, [markUnread, isRead]);
+
+  const allActions: {
+    [id in SwipeAction]: SlidingItemAction | undefined;
```
aeharding, line 228: `Record<SwipeAction, SlidingItemAction | undefined>`

aeharding:
What do you think about this?

```diff
--- a/src/features/settings/gestures/SwipeSettings.tsx
+++ b/src/features/settings/gestures/SwipeSettings.tsx
@@ -145,6 +145,7 @@ export default function SwipeSettings() {
 }

 const swipeIcons = {
+  [OSwipeActionAll.None]: " ",
   [OSwipeActionAll.Upvote]: arrowUpOutline,
   [OSwipeActionAll.Downvote]: arrowDownOutline,
   [OSwipeActionAll.Reply]: arrowUndoOutline,
```

image

aeharding:
Can we move Upvote above Downvote in the Setting Selector?
image

rsammelson:
> Can we move Upvote above Downvote in the Setting Selector?

It should already be like that, see the diff from my force push last night https://github.com/aeharding/voyager/compare/553969f2b5245bdab822ee36d4a0d444c20080e4..bef18183ce489791ad9350d9fb60968e1d6669d4. Is it not for you?

src/services/db.ts
rsammelson: I mean that now it's in the order: no information (none) -> some information (only the total) -> all information (separate scores). The default is the middle, and you can choose go up or down a level to see less or more. I can change it if you want.

src/features/settings/gestures/SwipeSettings.tsx
rsammelson: It turns out you can't super easily disable this. It turns on whenever you have an IonItem that is a button, because that is what iOS interfaces look like.

rsammelson: Never mind, I figured it out. Although it looks a little weird to me; if you go into iOS settings every single menu has a chevron.

rsammelson:
About the bug you mentioned, I can replicate it but I have not been able to figure out why it happens.

src/features/settings/gestures/SwipeSettings.tsx
aeharding: You should be able to disable with `detail={false}` or something. This is to match Apollo

rsammelson:
@aeharding can you take a look at #202? I would like to add share as a swipe option. Otherwise if this gets merged first I can add making it a swipe option to that PR.

aeharding:
> @aeharding can you take a look at #202? I would like to add share as a swipe option. Otherwise if this gets merged first I can add making it a swipe option to that PR.

Let's do share in a separate PR so this PR isn't blocked

src/features/settings/gestures/gesturesSlice.tsx
aeharding, line 1: rename to gestureSlice

src/features/shared/sliding/SlidingItem.tsx
aeharding: @rsammelson it still is black

rsammelson: I guess I accidentally marked this as done with the others, I'll fix this

rsammelson:
Most recent push was to update to the latest main, I wanted to ensure there were no problems from 9154e0c139f9f0fefebe since it sounded like it might touch something related to this.

aeharding:
I'm still a bit concerned about perf, but I haven't been able to get any conclusive results on perf changes between this and main.

@rsammelson Can you change the default far right swipe for posts from hide post to bookmark? I think this is getting pretty close now.

src/store.tsx
```diff
@@ -28,6 +31,7 @@ const store = configureStore({
     user: userSlice,
     inbox: inboxSlice,
     settings: settingsSlice,
+    gestures: gestureSlice,
```
aeharding, line 34: Probably change to "gesture" here too? Following "post" "comment" slice naming

src/services/db.ts
aeharding: Let's just keep it as is for now

src/features/settings/gestures/SwipeSettings.tsx
```diff
@@ -0,0 +1,233 @@
...
+  setInboxSwipeActionStart,
+  setAllSwipesToDefault,
+  setDisableLeftSwipes,
+  setDisableRightSwipes,
+} from "./gestureSlice";
+import { ActionCreatorWithPayload } from "@reduxjs/toolkit";
+import { Dictionary } from "lodash";
+import { useState } from "react";
+import { IonActionSheetCustomEvent, OverlayEventDetail } from "@ionic/core";
+import { ReactComponent as ShortSwipeSvg } from "./swipe_short.svg";
```
aeharding, line 34: I just realized that these files have underscores instead of camelCase too :)

src/store.tsx
rsammelson: Fixed. But why isn't it `setting: settingSlice`? :)

src/features/settings/gestures/SwipeSettings.tsx
rsammelson: This was to follow the naming in the public folder, it seemed like svg files used snake case.

aeharding: Oh, that's not right. We should be doing camel case (title case for react components).

src/store.tsx
aeharding: Technically, yeah :)

I suppose we could also rename to gestureSettings, but meh

src/features/settings/gestures/SwipeSettings.tsx
rsammelson: I can fix it tomorrow or feel free to push a fix

aeharding (approved):
Nice work, this is quite the enhancement.

rsammelson:
Thanks!

PR #519, opened 2023-07-24
New icon

fer0n:


fer0n:
@aeharding 

- I checked with [L1C4U5E@lemmy.world](https://lemm.ee/u/L1C4U5E@lemmy.world) and they're fine with the fediverse-star-sign being in here
- If you want to credit me somewhere, my Github profile probably makes the most sense (I'd prefer my GitHub username)
- Regarding the licence, do you need me to put something somewhere in a file or is it enough to tell you that it's all my stuff that I'm uploading here?
- Would you like the source file for all the variations etc in case something needs to be adjusted in the future (affinity designer)? If so, where do you want it?
- I think it'd make sense to let people know how they can customise the icon on iOS/Android. As mentioned before, on iOS there's [this shortcut](https://routinehub.co/shortcut/15817) that can be used (alongside other options) and I'm guessing there's a way to do that on Android as well. It would probably be best to do that alongside the URLs of the other contenders (maybe some more of the original submitted ones, I'm sure there are some in there that other people would prefer). There could also be a settings option that explains how and provides some alternative icons to choose from, although I'm not sure if the better way would be to wait for the potential native wrapper and do it there properly.

Let me know if something isn't quite right and I'll check it out :)

aeharding:
- Just added credits and license to the readme! Let me know if that's fine with you.
- For source files, can you attach them to this issue and/or a discussion? Then I can add them to a wiki page.
- I'll definitely mention that workaround for custom icon :)
- Lastly, I plan on working on an "about" page in the app which will be a good place for logo credit in-app! Does that sound good?

fer0n:
All of that is perfect, thanks :)

fer0n:
Here's the source file (Affinity Designer)
[Voyager icon.zip](https://github.com/aeharding/voyager/files/12147523/Voyager.icon.zip)

aeharding:
@fer0n please do!

fer0n:
#526

PR #526, opened 2023-07-25
Forcing high resolution icon for iOS

fer0n:


fer0n:
@aeharding I removed all the apple-touch-icons and added a 1024 one, that fixes it for me. Android isn't using these anyways. Afaik it's only being used for the iOS home screen and safari bookmark (the icons there). So there shouldn't be any unwanted side effects to this

aeharding:
We can just get rid of the unused icons, nbd.

PR #557, opened 2023-07-30
Revert "Revert "Add image thumbnails when possible""

aeharding:
Reverts aeharding/voyager#555


aeharding:
Let's try take two...

PR #563, opened 2023-07-30
iOS dark splash screen

fer0n:


aeharding:
Thanks for doing that, works perfect!

Also, I don't know if you noticed, but the splash is pinned to the bottom of the screen. Just something I did to get things working.

I don't know if there's potentially a better way of doing that so it would look great on a wide variety of devices - lmk if you have any ideas!

aeharding:
I trust your judgement on that @fer0n, either way seems fine. Blank is definitely easier... :)

fer0n:
I went for the skeleton variant in #569. The native navigation bar didn't align with the one in voyager, but the tab bar did, so I added that and the color for light and pure black should match. The non-pure black one doesn't match, not sure if that's possible somehow and/or if there's an overall better solution for this. Either way, I think it's less distracting this way and feels a bit faster, what do you think?

PR #569, opened 2023-08-01
skeleton splash screen

fer0n:


aeharding:
Just pulled down to test, doesn't look quite right for some reason:

IMG_3241

fer0n:
It looked correct for me, I'll have another look

fer0n:
Can you test it again? It should look better now.

aeharding:
Looks good!

PR #605, opened 2023-08-06
Fix android back button closing app when gallery is open

### LINKED ISSUE #601: Pressing Android back button from an image exits app
This bug seems to affect only the Android version from the play store and not the PWA version on my device. This basically renders the app useless for me.

aeharding:
Resolves #601


eeshankeni:
Does this work only on the Android app version? Afaik this [event handler](https://ionicframework.com/docs/developing/hardware-back-button#hardware-back-button-in-a-browser) does not fire on browsers. Been having the same issue with my app while trying to use [react-spring-lightbox](https://github.com/tim-soft/react-spring-lightbox)

PR #627, opened 2023-08-09
feat: cake day 🍰

sharunkumar:
addresses #59

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,8 +26,8 @@ jobs:
       - name: 📦 Install dependencies
         run: pnpm install  --frozen-lockfile
 
-      # - name: 🧪 Test project
-      #   run: pnpm test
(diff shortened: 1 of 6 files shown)


aeharding:
Thanks for the PR! I'll check it out shortly

sharunkumar:
@aeharding so with the current patch, people who have just created an account < 24h also shows as cake day. is this the way its supposed to work?

rsammelson:
Will this show the icon at different times for people in different time zones?

sharunkumar:
> Will this show the icon at different times for people in different time zones?

time is not taken into account when comparing today's date and the published date. so I think it shouldn't be an issue, since both dates are converted to local time? :thinking:

aeharding:
Logic should now take everything into account - and also not show cake day for brand new users. :)

I added some tests too in case more edge cases come up!

PR #673, opened 2023-08-16
feat: Distinguish Post Contexts For Blurring Logic

ezfe:
I assume I've violated some project conventions with this PR, so please let me know what your thoughts are and if I should be moving anything around.

This changes the blurring logic from a previously Always/Never option to introduce a third option. Now the user can choose to unblur only within NSFW contexts and default to blurring in Home, Popular, All, and any SFW community.

I was a little unsure on some of the labeling, and the type I added (postContext) I wasn't entirely sure where to drop that.

### DIFF
--- a/src/features/feed/PostCommentFeed.tsx
+++ b/src/features/feed/PostCommentFeed.tsx
@@ -9,6 +9,7 @@ import { receivedComments } from "../comment/commentSlice";
 import Post from "../post/inFeed/Post";
 import CommentHr from "../comment/CommentHr";
 import { FeedContext } from "./FeedContext";
+import { PostContext } from "../../helpers/postContext";
 
 const thickBorderCss = css`
   border-bottom: 8px solid var(--thick-separator-color);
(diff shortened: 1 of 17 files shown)


ezfe:
@aeharding I've left this as a Draft to represent that while this is fully functional I'm not confident in it's ability to be merged until it's been fully vetted

aeharding:
Hi! This is an interesting idea. So the idea is you would only see NSFW unblurred in feed if you're browsing a NSFW community?

ezfe:
Yup - the context of the post allows the check to be based on the feed the post is in.

If the user is in an NSFW community then the post is unblurred and if the user is in a SFW community or an aggregated feed like Popular then it remains blurred.

aeharding:
@ezfe Is this something you would still like added to Voyager? I noticed that there are quite a few conflicts and it is still marked as a draft.

Btw, I made #1382 to make sure I don't lose track of this.

ezfe:
I feel like it would be a good idea, but I have not been using Voyager recently so take that as you will. I don't have a development environment to resolve the conflicts with at this time unfortunately.

aeharding:
Okay, thanks! I am going to close this for now then, since there is the issue I created to track.

PR #674, opened 2023-08-16
Added alternative icons

fer0n:
I just added them to the iOS Assets, let me know if you need some adjustments. Also, here's the updated source file
[Voyager icon.zip](https://github.com/aeharding/voyager/files/12356136/Voyager.icon.zip)


aeharding:
Thanks very much @fer0n! I've integrated these into #675 :)

PR #704, opened 2023-08-21
fix some images not being available in the file picker

sharunkumar:
Wasn't able to select some images to use in posts/comments. This change fixes that

Screenshot_20230821_130130_Files_2

Note: looks like lemmy file input uses both `images/*, video/*`, but since uploads are routed through the voyager instance, I am not sure how it would impact it. so I have omitted video/* from this patch

Screenshot_20230821_131953


aeharding:
Thanks!

PR #719, opened 2023-08-31
translucent tab bar in ios

rahulakrishna:
- add translucent property to `<IonTabBar>`
- fix the style in `index.css`

### Dark Mode (with pure dark disabled)
image

### Light Mode
image

### Pure Dark Mode
image


sharunkumar:
does this not work in the PWA? 🤔

PR #750, opened 2023-10-03
Add lemmy v0.19 authentication support

aeharding:
NOTE: This just fixes auth for lemmy v0.19.

One month after lemmy v0.19 is released, a new version of Voyager will be released that phases out support for v0.18.

This will allow Voyager to adopt new v0.19 features.

NOTE 2: A new, breaking change of Voyager (Voyager v2) will likely be released sooner than that. Since the native clients can send a custom `Cookie` header, we can upgrade lemmy-js-client sooner than that. The web version will remain stuck of Voyager v1 until most Lemmy servers have a chance to upgrade to v0.19 (one month after v0.19 release)

NOTE3: Phase 2 PR is coming soon.

### DIFF
--- a/server.mjs
+++ b/server.mjs
@@ -122,7 +122,8 @@ app.use(
         req.path === "pictrs/image" &&
         req.query?.auth
       ) {
-        clientReq.setHeader("cookie", `jwt=${req.query.auth}`);
+        clientReq.setHeader("cookie", `jwt=${req.query.auth}`); // lemmy <=v0.18
+        clientReq.setHeader("Authorization", `Bearer ${req.query.auth}`); // lemmy >=v0.19
         delete req.query.auth;
       }
     },
--- a/src/features/auth/Login.tsx
+++ b/src/features/auth/Login.tsx
@@ -171,12 +171,7 @@ export default function Login({
 
     try {
       await dispatch(
-        login(
-          getClient(server ?? customServerHostname),
-          username,
-          password,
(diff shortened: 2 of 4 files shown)


phiresky:
Hey. quick question about this. even with 0.19 you still get the jwt token in the login body json response.

And you don't need to set a cookie header, you can also set a ```Authorization: `Bearer ${jwt}` ``` header, which should work fine with the fetch api (https://stackoverflow.com/a/48943220/2639190 ). Did you consider this when saying you need to break backwards compat?

aeharding:
Lemmy 0.18 doesn’t support the authorization header afaik. Only the cookie header (or auth url parameter)

phiresky:
yes but you can still send both right? should be independent of other new features. I mean make sense to drop support at some point just wondering about your reasoning and if there's any incompatibility on Lemmy s side

aeharding:
Yes, I can send both - However, lemmy-js-client v0.19 removes support for `auth` URL param. Which means that the only way to authenticate with lemmy v0.18 using lemmy-js-client v0.19 is to use the `cookie` header (which is incompatible with web-based clients)

That's why this PR keeps lemmy-js-client v0.18 - as a stopgap to allow compatibility with lemmy v0.18 and v0.19 for web-based clients. The only downside is since lemmy-js-client is stuck on v0.18, I can't take advantage of any new v0.19 features.

PR #571, opened 2023-08-01
Show Imgur and other videos in the Feed and the Post view

carloscheddar:
This fix allows posts that have an `embed_video_url` to be able to load on the app instead of needing to visit the url. Since the `embed_video_url` returns a `.mp4` link this allows us to load videos from Imgur and other gif/video services using the existing `Video` component.

Other sites like youtube require an embed instead which I plan to look into in another PR if no one has gotten to that yet.

PR #791, opened 2023-10-17
Show more videos and consolidate media logic

carloscheddar:
Given a Lemmy post containing **media** this will try to display the content inside the app. This uses the existing image and video renderers and adds a new `iframe` for video embeds.

### Features
- Render videos via `embed_video_url`
- Automatically transform media urls into one that we can render or embed (`.gifv` -> `.mp4`, etc.) 
- Allow rendering `.mov` urls as videos
- Add error handling for images and videos as well as fallbacks
- Consolidate media logic into a single `Media` component used by both `PostDetail` and `LargePost`
- Add `isUrlMedia` to easily check for media instead of manually checking for image, video or embed
- Add tests for the url helpers. Helps with the url transformations since these can increase

### Fallbacks
- When an image `url` fails to load try rendering the `thumbnail` instead
- When a video fails to load via `embed_video_url`, try rendering the `url`
- When everything fails to load, display the `url`

### DIFF
--- a/src/features/comment/links/LinkPreview.tsx
+++ b/src/features/comment/links/LinkPreview.tsx
@@ -11,8 +11,8 @@ import {
 import { css } from "@emotion/react";
 import { getImageSrc } from "../../../services/lemmy";
 import { ReactNode, useMemo } from "react";
-import { isUrlImage } from "../../../helpers/lemmy";
 import useLemmyUrlHandler from "../../shared/useLemmyUrlHandler";
+import { isUrlImage } from "../../../helpers/url";
 
 const shared = css`
   width: 30px;
--- a/src/features/gallery/GalleryImg.tsx
+++ b/src/features/gallery/GalleryImg.tsx
@@ -1,4 +1,10 @@
-import React, { FocusEvent, KeyboardEvent, useContext, useRef } from "react";
+import React, {
+  FocusEvent,
+  KeyboardEvent,
+  ReactEventHandler,
+  useContext,
+  useRef,
+} from "react";
 import "photoswipe/dist/photoswipe.css";
 import { PostView } from "lemmy-js-client";
 import { GalleryContext } from "./GalleryProvider";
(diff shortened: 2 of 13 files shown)


aeharding:
Thanks for the PR! I'll take a look as soon as I can

aeharding:
Hi there, thanks again for the PR!

So, I have a couple concerns about the approach taken in this PR.

1. We should never load iframes to content such as YouTube in the app without the user requesting to do so (for example, by tapping a static preview picture). This is the approach that Mastodon takes, and its best for security and privacy.
2. Ideally tapping the preview picture would bring up the lightbox with the same behavior that images have. This is what Apollo does. Note, Apollo doesn't load iframes in the app. It somehow extracts the Youtube video data and puts it into its own player. This is a much better experience, and better for privacy reasons, but may be hard to implement.

Overall this isn't the approach I'd like to take for Voyager, but I do appreciate your efforts here! Let me know if you have any questions on the above.

carloscheddar:
@aeharding I think your concerns about iframes are valid I even forgot to add the `sandbox` attribute! I'll remove the embed logic from this PR and revisit that in the future.

I think the biggest benefit of this PR is the consolidation and error handling for images and videos as well as being able to render video files from the `embed_video_url` that Lemmy provides.

Let me know if you have any comments on those.

aeharding:
There was a previous PR playing with support for embed_video_url in the past, see here: https://github.com/aeharding/voyager/pull/571/files

The problem with that at the time was a lot of embed_video_url returned by lemmy is invalid, which is why Voyager doesn't support it at the moment. I don't remember why exactly, maybe browsers had trouble handling the media files. Might be worth checking again though

carloscheddar:
> There was a previous PR playing with support for embed_video_url in the past, see here: https://github.com/aeharding/voyager/pull/571/files
> 
> The problem with that at the time was a lot of embed_video_url returned by lemmy is invalid, which is why Voyager doesn't support it at the moment. I don't remember why exactly, maybe browsers had trouble handling the media files. Might be worth checking again though

Yup that was my previous PR, this fixes that issue thanks to the fallback logic via the onError events. I just went a little overboard and tried to add embeds too :laughing: .

aeharding:
I see, thanks!

aeharding:
I am closing this because it has become stale and there have been various fixes made to the video player over the past month or so. Thank you though for the PR!

Issue #108, opened 2023-07-01
Support lemmy markdown image syntax ![](imagefile.jpg)


Barusu-rem:
Is it possible to support lemmy image syntax?

Example

`![](https://i.imgur.com/fghSMUn.png)`

![](https://i.imgur.com/fghSMUn.png)

aeharding:
Hi!

It is an intentional choice to hide all images in comments.

But we could support an option in settings to enable (off by default).

PR #827, opened 2023-10-24
feat: add setting to render images in comments

sharunkumar:
Resolves https://github.com/aeharding/voyager/issues/108

Disabled by default: https://github.com/aeharding/voyager/issues/108#issuecomment-1615435828

Also in this PR:
- Render compact post settings only when relevant setting is selected


src/features/settings/general/comments/RenderCommentImages.tsx
```diff
@@ -0,0 +1,22 @@
...
+export default function RenderCommentImages() {
+  const dispatch = useAppDispatch();
+  const { renderCommentImages } = useAppSelector(
+    // this needs a better naming
+    (state) => state.settings.general.comments,
+  );
+
+  return (
+    <InsetIonItem>
+      <IonLabel>Render Comment Images</IonLabel>
```
aeharding, line 15: Lets rename everything from "render" comment images to "show" comment images

src/features/settings/appearance/CompactSettings.tsx
```diff
@@ -18,6 +18,12 @@ export default function CompactSettings() {
   const { thumbnailsPosition, showVotingButtons, thumbnailSize } =
     useAppSelector((state) => state.settings.appearance.compact);
 
+  const postsAppearanceType = useAppSelector(
+    (state) => state.settings.appearance.posts.type,
+  );
+
+  if (!(postsAppearanceType === "compact")) return <></>;
```
aeharding, line 25: Lets still render if compact mode - its what apollo does :)

src/features/comment/CommentMarkdown.tsx
```diff
@@ -6,19 +7,27 @@ interface CommentMarkdownProps {
 }
 
 export default function CommentMarkdown({ children }: CommentMarkdownProps) {
+  const { renderCommentImages } = useAppSelector(
+    (state) => state.settings.general.comments,
+  );
+
   return (
     <Markdown
```
aeharding, line 15: Can we make this Markdown a StyledMarkdown and add css to target `img` inside with a max-height? Perhaps 200px will do.

aeharding:
One other thing - we should probably edit `CommentLinks` and filter out `type = 'image'` if you have show comment images turned on.

src/features/settings/appearance/CompactSettings.tsx
sharunkumar: You mean, when compact mode is off?

aeharding: Yeah, I would like to keep the previous behavior where the compact settings show all times.

There will probably be a feature in the future where you can change compact to large post mode in the feed.

PR #828, opened 2023-10-24
feat: add a button to fetch post titles

### LINKED ISSUE #817: Automatically fill article/video title when sharing a link
When sharing an article or a video, usually the post title is the same title that the article's. So I'm going back and forth twice to copy the title as well.

I've made an [iOS Shortcut](https://routinehub.co/shortcut/16651/) basically just for this, but having it in the App would be really nice.

sharunkumar:
should resolve #817

~~currently uses proxying to fetch in case of PWA, due to CORS issues when doing it directly~~

image

### DIFF
--- a/src/features/post/new/PostEditorRoot.tsx
+++ b/src/features/post/new/PostEditorRoot.tsx
@@ -371,6 +371,7 @@ export default function PostEditorRoot({
             <IonItem>
               <IonInputTitle
                 value={title}
+                clearInput
                 onIonInput={(e) => setTitle(e.detail.value ?? "")}
                 placeholder="Title"
                 counter
--- a/src/features/post/new/PostEditorRoot.tsx
+++ b/src/features/post/new/PostEditorRoot.tsx
@@ -421,28 +421,24 @@ export default function PostEditorRoot({
                   value={url}
                   onIonInput={(e) => setUrl(e.detail.value ?? "")}
                 />
-                {!!url && isValidUrl(url) && (
+                {!!jwt && !!url && isValidUrl(url) && (
                   <IonButton
-                    hidden={!jwt}
                     onClick={(e) => {
                       e.preventDefault();
-                      async function fetchData() {
-                        if (!jwt) return;
-                        const data = await getSiteMetadata(
-                          instanceUrl,
-                          jwt,
-                          url,
-                        );
-                        if (!data?.title) {
-                          return presentToast({
-                            message: "Unable to fetch title",
-                            color: "danger",
-                          });
-                        }
-                        setTitle(data.title);
+                      function toast() {
+                        presentToast({
+                          message: "Unable to fetch title",
+                          color: "danger",
+                        });
                       }
-
-                      fetchData();
+                      getSiteMetadata(url, instanceUrl, jwt)
+                        .then((data) => {
+                          if (!data.title) {
+                            return toast();
+                          }
+                          setTitle(data.title);
+                        })
+                        .catch(toast);
                     }}
                   >
                     FETCH TITLE
--- a/src/features/post/new/PostEditorRoot.tsx
+++ b/src/features/post/new/PostEditorRoot.tsx
@@ -325,6 +325,22 @@ export default function PostEditorRoot({
     setPhotoPreviewURL(undefined);
   }
 
+  async function fetchPostTitle() {
+    try {
+      const { metadata } = await client.getSiteMetadata({
+        url,
+      });
+
+      if (metadata.title) {
+        setTitle(metadata.title);
+      } else {
+        presentToast(problemFetchingTitle);
+      }
+    } catch (error) {
+      presentToast(problemFetchingTitle);
+    }
+  }
+
   return (
     <>
       <IonHeader>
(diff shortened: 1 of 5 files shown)


aeharding:
Thanks for the PR! Why not use [getSiteMetadata](https://join-lemmy.org/api/classes/LemmyHttp.html#getSiteMetadata) like lemmy-ui, though?

sharunkumar:
> Thanks for the PR! Why not use [getSiteMetadata](https://join-lemmy.org/api/classes/LemmyHttp.html#getSiteMetadata) like lemmy-ui, though?

was not aware of this, will check it out :eyes:

src/features/post/new/PostEditorRoot.tsx
```diff
@@ -419,6 +421,26 @@ export default function PostEditorRoot({
                   value={url}
                   onIonInput={(e) => setUrl(e.detail.value ?? "")}
                 />
+                {!!jwt && !!url && isValidUrl(url) && (
+                  <IonButton
+                    onClick={(e) => {
```
aeharding, line 426: Lets use async/await for this function

src/features/post/new/PostEditorRoot.tsx
```diff
@@ -419,6 +421,26 @@ export default function PostEditorRoot({
...
+                      e.preventDefault();
+                      function toast() {
+                        presentToast({
+                          message: "Unable to fetch title",
+                          color: "danger",
+                        });
+                      }
+                      getSiteMetadata(url, instanceUrl, jwt)
```
aeharding, line 434: instead of adding this function to lemmy.ts, add `const client = useClient()` to the top of the component and then call `client.getSiteMetadata()` directly (`const { metadata } = await client.getSiteMetadata()`)

something like-

```ts
let metadata;

try {
  ({ metadata } = await client.getSiteMetadata())
} catch(error) {
  presentToast(siteMetadataError)
  throw error;
}

if (!data.title) {
  presentToast(siteMetadataError)
  return
}

setTitle(metadata.title)
```

src/features/post/new/PostEditorRoot.tsx
```diff
@@ -419,6 +421,26 @@ export default function PostEditorRoot({
                   value={url}
                   onIonInput={(e) => setUrl(e.detail.value ?? "")}
                 />
+                {!!jwt && !!url && isValidUrl(url) && (
+                  <IonButton
+                    onClick={(e) => {
+                      e.preventDefault();
+                      function toast() {
+                        presentToast({
```
aeharding, line 429: We can add this object to `toastMessages.ts`

src/features/post/new/PostEditorRoot.tsx
```diff
@@ -419,6 +421,26 @@ export default function PostEditorRoot({
...
+                        });
+                      }
+                      getSiteMetadata(url, instanceUrl, jwt)
+                        .then((data) =>
+                          data.title ? setTitle(data.title) : toast(),
+                        )
+                        .catch(toast);
+                    }}
+                  >
+                    FETCH TITLE
```
aeharding, line 441: Can we do something shorter and more subtle, maybe remove the bg too? Maybe there's an appropriate icon we can use (https://ionic.io/ionicons)

sharunkumar: I was initially experimenting with some of the icons but wasn't sure which one to go with, so I just went with a button for now. open to suggestions for the icon for this button :v:

aeharding:
@sharunkumar I moved around the field a bit. Now, it displays "autofill" next to the title box after you've entered a URL, but not if you've already typed in the title field.

Let me know what you think.

sharunkumar:
> @sharunkumar I moved around the field a bit. Now, it displays "autofill" next to the title box after you've entered a URL, but not if you've already typed in the title field.
> 
> Let me know what you think.

looks perfect :+1:

fer0n:
Thanks for working on this :) 

Regarding the implementation: there's an argument to be made that it should auto fill whenever the title is empty and a URL is entered. The button to trigger the auto fill could still be there, but wouldn't have to (makes the UI a bit cleaner without, but being deliberate has also advantages).

I'd personally go with auto fill an empty title (show a spinner in the title field while loading) and otherwise do nothing, but that's up to you :)

sharunkumar:
> it should auto fill whenever the title is empty and a URL is entered

this is there in the latest commit made by @aeharding

fer0n:
Ah, nice. Apologies for the comment then.

src/features/post/new/PostEditorRoot.tsx
```diff
@@ -77,6 +80,8 @@ const HiddenInput = styled.input`
 
 type PostType = "photo" | "link" | "text";
 
+const MAX_TITLE_LENGTH = 200;
```
sharunkumar, line 83: @aeharding is this a limit imposed by the back end? :thinking:

aeharding: Yeah - it's hardcoded in Lemmy last I checked (https://github.com/iav/lemmy/commit/65a6e7731c632313f539d232dec4f6f6c8a49084)

aeharding:
> Ah, nice. Apologies for the comment then.

It doesn't *automatically* fetch the post title - you still need to press the button.

I thought about that, however in some cases the request took an abnormal amount of time, so I opted not to. Once we have this in the wild we can totally revisit this though, it would be a very small change!

aeharding:
Thanks for this!

Issue #608, opened 2023-08-06
Add support for opening Lemmy URLs


gobbledygoober:
Hoping for easier ways to use Voyager to open Lemmy URLs (from outside the app). I'm sure contributors are already thinking about this, but I didn't see this captured in an existing open/closed issue. 

Use cases: 
1. A Lemmy post URL is shared with you.
1. Comparing/testing how posts are rendered in the various emerging Lemmy clients.
1. Want to switch to Voyager after opening a post in a mobile browser.

As of 1.4.0, I believe the best workflow is to use the (pretty great) search feature and use terms that appear in tht desired post.  

I am primarily using the iOS app now, so some of these are iOS-centric, but I'm sure some are valid for PWA and Android. This is an attempt at an somewhat exhaustive list of potential implementations (inspired by [this Reddit post](https://www.reddit.com/r/apolloapp/comments/9xw2e3/how_to_open_a_reddit_link_in_apollo/) that helped refresh my memory).

I imagine that some of these are trickier than others because Lemmy URLs aren't all from a single domain. I am no developer, so the compatibility column is my best guess. 

All of these describe what to do with a Lemmy post URL:

| # | Method | Apollo Parity | App Compatibility | Notes |
| --------| --------------- | --------------- | --------------- | -------------- |
| 1 | ✅ DONE ~~Paste into Voyager search~~ | Maybe? | iOS, Android, PWA | Bean detects when a URL is pasted into search, but doesn't show any results when a lemmy.world post URL was used. <br><br>Last week I swear this worked on one of the other iOS clients (Memmy, Mlem, Avelon, Lemmios, Bean, Lemmotif, Liftoff!, Thunder, Lemur) ... but trying again with the latest versions as of this post did not work for any of them. | 
| 2 | ❌ WONTFIX ~~Prompt to open from clipboard on Voyager launch~~ | Yes | iOS, Android, maybe PWA? | Long press the URL. Copy. Open Voyager. Click yes/OK when it prompts to open the link. | 
| 3 |  ✅ DONE iOS ~~"Open in Voyager" from Share Sheet~~ | Yes | iOS, maybe Android | Believe this was implemented as an iOS Shortcut that you could install from Apollo. I am less familiar with Android but perhaps Tasker could facilitate this. 
[...]

PR #829, opened 2023-10-24
Add URL searching, refactor linking, fix crash

aeharding:
Related #608


gobbledygoober:
@aeharding Is there a reason that this would not work with kbin URLs? Searching "https://kbin.social/m/AccidentalRenaissance" did not turn up a community result, but searching "AccidentalRenaissance" alone does.

aeharding:
Right now it’s just implemented for Lemmy style links. Could in the future add support for other styles like mastodon too, but wouldn’t be trivial.

PR #848, opened 2023-10-28
Improve performance of comment collapse

aeharding:
Previously, all comments were forced to run a render cycle due to this useSelector(). With this fix, the useSelector() will only trigger a render of the comment that was collapsed.

This appears to have a noticeable performance improvement, especially on low-end devices.


SrivatsanSenthilkumar:
Thanks for this. You have no idea how greatly a simple thing like this has improved the user experience.

Issue #381, opened 2023-07-11
Old account identifier 👵👴


kerosyn:
Related to #17

I think it would be fun to have the option of identifying older accounts. And by old, I mean pre-dating the Reddit drama that caused the influx of new Lemmy users. Whether that was the date the api changes were announced, the 3pa last day, or the start of the new api policy.

It would just be fun way to see veterans of the platform at a glance.

aeharding:
Haha. That's a fun idea. It does have the idea to be gamed though.

kerosyn:
> Haha. That's a fun idea. It does have the potential to be gamed though.

It’s your call as it’s your project, but outside of the bot accounts that many of which were deleted, I don’t see how it can be gamed. Also only Voyager users will see it unless other clients implement it.

The idea was inspired by the game series Battlefield’s old veteran status of owners of the older games.

aeharding:
I believe you can technically set your account age to anything if you have access to the instance DB.

But again, I think the possibility of people gaming this is fairly low :)

aeharding:
I've thought about this more, and my main concern is that users may not want their account tagged with this badge. Unlike the new account badge, where if you wait long enough it'll go away, this badge does not.

For that reason I don't think I am going to implement this.

That being said, I am absolutely open to badges in general and thank you for bringing this idea up!!

PR #917, opened 2023-11-10
easter egg: detect users from the future

### LINKED ISSUE #914: Need to handle invalid dates
Screenshot_20231109_194920_Chrome

Screenshot_20231109_195030_Chrome

I can work on a PR for this later

> https://github.com/aeharding/voyager/issues/381#issuecomment-1636867004

sharunkumar:
Resolves #914

Related: https://github.com/aeharding/voyager/issues/381#issuecomment-1636867004

image


aeharding:
Hi, thanks for the PR! I think we should just hide the new account indicator if the account created date is in the future. Otherwise it creates an incentive for people to mess with their DB and break account created dates for other apps

sharunkumar:
But the possibility of abusing this is low as you had mentioned it earlier. Only instance owners have direct access to their DBs.

Either way we already know that this is already being done in the wild, and apps would have to handle this edge case anyway. Which is why I thought it would be a funny easter egg when you come across such a user.

aeharding:
> But the possibility of abusing this is low as you had mentioned it earlier. Only instance owners have direct access to their DBs.

Perhaps, but I still don't want to create any incentive for people to do that :)

It would definitely be a fun easter egg, but in this case, I think it's best to ignore. Thanks again for the PR but I'm closing in favor of #925 :)

sharunkumar:
@aeharding would you consider it if the changes were behind a hidden feature flag (disabled by default) which I could just flip downstream?

PR #918, opened 2023-11-10
Add in-app purchases for tips

aeharding:


SrivatsanSenthilkumar:
Out of curiosity, will apple/google receive a cut of this donation. If yes, could you put a message that people can pay using github or librepay so more of the money people are tipping/donating reaches your hands. This is a great idea though and thunderbird got a great amount of donations by having in app tips/donations messages(which were out of the way and not annoying popups) and option to tip/donate in app as well

aeharding:
> If yes, could you put a message that people can pay using github or librepay so more of the money people are tipping/donating reaches your hands. 

unfortunately, I don’t think the App Store allows advertising paying externally :/ they want their 15%

SrivatsanSenthilkumar:
> > If yes, could you put a message that people can pay using github or librepay so more of the money people are tipping/donating reaches your hands. 
> 
> unfortunately, I don’t think the App Store allows advertising paying externally :/ they want their 15%

Wasn't there some law passed after the apple/Google vs epic case where it was mandated that apple and Google have to allow 3rd payment portals for in app purchases.

aeharding:
I don't think so. At least, not yet?

image

SrivatsanSenthilkumar:
> I don't think so. At least, not yet?
> 
> image
> 

Sadge. At least the pwa and f-droid builds can like the GitHub. Hopefully this will get more people to donate and reward you for all the hard work

PR #924, opened 2023-11-11
feat: spoiler support

### LINKED ISSUE #9: Implement spoiler tag support
```
:::spoiler

(stuff here)

:::
```

is supported by lemmy. It appears to be custom spoiler markup? So might need to implement our own markdown plugin. :/

DakshG07:
Resolves #9. In its current state, most likely buggy (there isn't much in the way of error handling right now, so any improper spoiler tags will most likely crash Voyager). 

~~Opening spoilers also collapse the comment/post at the moment. Not quite sure how to fix that.~~ This is now fixed.


aeharding:
Hi, thanks for the pr! This is moving in the right direction.

 1. We will definitely need tests for this code to show markdown input produces expected HTML output in various scenarios
 2. The `spoiler://` seems a bit too hacky and goes against the grain of remark/rehype. We should be able to output HTML detail and summary tags directly
 3. This plugin feels like maybe it should live in a separate repository. I think many other projects would find it useful, and then everyone can collaborate and improve in one place :)

DakshG07:
I agree with the first point, this is really fragile at the moment and probably needs much more testing.

As for 2, I'll have to look more into the capabilities of `rehype`. If I can set the `onClick` handler with `rehype`, then yes, we should be good there.

For 3, I have no problem putting this in its own repository. The `remark` and `rehype` plugin method used here isn't exactly the ideal method of implementing this, but some other projects may find it useful (it is quite a specific project though, as Lemmy uses its own spoiler syntax).

aeharding:
Cool. Feel free to keep the PR open if you'd like for tinkering/prototyping until a repo is made!

> If I can set the onClick handler with rehype, then yes, we should be good there.

Yeah, I'm not entirely sure how the internals work, but in terms of an external API surface I was hoping we could just add another [component to ReactMarkdown here](https://github.com/aeharding/voyager/blob/a2bce4d45122579f908c60fb362322b6df340641/src/features/shared/Markdown.tsx#L67) - something like

```tsx
        summary: (props) => (
          <summary onClick={(e) => e.stopPropagation()} {...props} />
        ),
```

DakshG07:
At first I wanted to do this, but the documentation wasn't giving me any ideas on how to implement this. That was before I'd started working on the `rehype` plugin. After working with `hast`, it seems that the components just rely on the `tagName` of the `Element`, so this can actually be implemented quite simply. Typescript is complaining about this however (I'm guessing since `spoiler` isn't technically a valid field), so I'll have to try to fix that.

DakshG07:
Seems concerns 2 and 3 have been met! As for testing, I wrote a couple (and actually discovered a pretty big bug in the process), but I'm feeling a bit lazy and'll probably write the rest of them later. The code for the spoiler plugin has been moved [here](https://github.com/DakshG07/remark-lemmy-spoiler), and it now creates a custom `<spoiler>` tag, which allows us to render it however we like.

_P.S. If you know how to get around using [this workaround](https://github.com/DakshG07/voyager/blob/6443fc1d65932bece43639f25f70e0c4c32c45cf/src/features/shared/Markdown.tsx#L77), that would be greatly appreciated!_

80avin:
@DakshG07  Regarding your workaround, you have to extend an interface defined in some other library.

I think this stack overflow answer explains in a more concise manner than official TS documentation 

https://stackoverflow.com/a/44828876/11218031

aeharding:
Thanks for your work on this, and pulling into a separate repo.

I'm a bit worried by two things: Testing, and performance (regarding [your PSA on the repo](https://github.com/DakshG07/remark-lemmy-spoiler#psa) about regexes and combing through the tree in a hacky manner). Performance is a concern because this is a significant portion of render time with Voyager. Testing is a concern because I don't want breaking other markdown syntax.

So this is a cool POC, and hopefully we can continue to finesse to get it into a state where it will be ready to be added to Voyager :)

(Also, is it possible for the plugin to not create a custom `spoiler` tag, but instead use `detail` and `summary` directly?)

When I have time, I'll play with this more! Thanks again

aeharding:
I am closing this in favor of https://github.com/aeharding/voyager/pull/1270. Thanks @DakshG07 for your help on this!

PR #933, opened 2023-11-14
feat: remember previously used links for migration

sharunkumar:
Screenshot_20231114_123121_Chrome


src/pages/settings/RedditDataMigratePage.tsx
```diff
@@ -108,7 +146,19 @@ export default function RedditMigratePage() {
       <IonHeader>
         <IonToolbar>
           <IonButtons slot="start">
-            <IonBackButton defaultHref="/settings" text="Settings" />
+            {subs ? (
+              <IonButton
+                onClick={(e) => {
```
aeharding, line 151: This should probably navigate to a separate page if we're going to allow navigating back.

The migrate page is quite terse/simple, so this will require a bit of refactor if we want to increase the complexity with saving migrate links.

src/features/community/migrationSlice.ts
```diff
@@ -0,0 +1,70 @@
...
+  reducers: {
+    setMigrationLinks: (state, action: PayloadAction<string[]>) => {
+      state.links = action.payload;
+    },
+  },
+});
+
+export default migrationSlice.reducer;
+
+export const addMigrationLink =
```
aeharding, line 26: add/reset can be reducers.

sharunkumar: I made the changes based on `src/features/community/communitySlice.ts` 

should I move `addMigrationLink`,`resetMigrationLinks`, including the db commands into the reducer section?

aeharding: I see. Yeah those favorites should be reducers too. Take a look at settingsSlice

sharunkumar: I'll work on these suggestions next week. Just keeping up with the conflicts for now

sharunkumar: @aeharding I've updated this, let me know if there's anything else in this context

aeharding:
Hi @sharunkumar!

Thanks for your patience. I just had some time to take a look. I refactored a few things:

 1. Split up views into separate pages so swiping works properly
 2. Added swipe to forget, so you can only clear what you want. It also cleans up the UI a little bit and makes it similar to the filters/blocks UI.
 3. The paste for migrate link is now in an alert. It's a few more taps, but now that it is remembered, I don't think it's that big a deal.

If you could try it out, and let me know what you think, I would appreciate it and we can finally get this merged.

Thanks!

aeharding (approved):

sharunkumar:
Looks good! Thanks for the update :D

Issue #935, opened 2023-11-15
Crash IOS 17.2 beta 3


Welch22:
### Crash description

 

Trying to load a post or comment reply resulted in an immediate crash. iOS 17.2 beta 3 IPhone 13 Pro Max. 

### Device and app metadata

  - window.location.href: `capacitor://localhost/`
  - react-router location.pathname: `/posts/lemm.ee/c/marvelstudios@lemmy.world/comments/14930248`
  - Logged in? `true`
  - Native app? `true`
  - Installed to home screen? `true`
  - Voyager version: `1.23.2`
  - BUILD_FOSS_ONLY: `false`
  - User agent: `Mozilla/5.0 (iPhone; CPU iPhone OS 17_2 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148`

### Crash data

Error: `TypeError: undefined is not an object (evaluating 'e.linked.filter')`

#### Stack trace

```
@capacitor://localhost/index.js:216:190067
d@capacitor://localhost/index.js:209:48939
@capacitor://localhost/index.js:209:50120
d@capacitor://localhost/index.js:209:48939
l@capacitor://localhost/index.js:232:439
useSyncExternalStore@capacitor://localhost/index.js:40:70954
@capacitor://localhost/index.js:232:717
@capacitor://localhost/index.js:232:1664
om@capacitor://localhost/index.js:248:111873
Af@capacitor://localhost/index.js:248:183401
B5@capacitor://localhost/index.js:40:63935
u7@capacitor://localhost/index.js:40:116540
l7@capacitor://localhost/index.js:40:112269
DK@capacitor://localhost/index.js:40:112197
rb@capacitor://localhost/index.js:40:112050
sk@capacitor://localhost/index.js:40:108412
a8@capacitor://localhost/index.js:40:109219
fl@capacitor://localhost/index.js:40:47676
@capacitor://localhost/index.js:40:106742
```

aeharding:
Thanks! Does this happen if you force restart the app? If you clear the app data, does the problem go away?

Welch22:
I ended up clearing the app data and that solved it for now. Force restarting didn’t fix it tho.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

aeharding:
OK, thanks! I'll have a fix that should resolve this, although it's still not completely clear the cause.

PR #936, opened 2023-11-15
Fix crash receiving linked instances (#935)

aeharding:


aeharding:
Not sure of the root problem - it could just be an iOS beta bug. But the "fix" is quite small.

Issue #621, opened 2023-08-07
Delete + Cancel stay on screen after posting comments (web app)


nightgolfer:
When I'm writing a comment, I have a button for Cancel top left, and Post top right.

After finishing the comment and clicking Post, a green message pops up bottom screen saying "Comment posted!". Underneath that are two new "buttons", Delete and Cancel".

The screen stays on the comment input "page", with Cancel and Post still at the top of the screen; I'm not sent back to the post I was commenting on, and the two new Delete and Cancel options at the bottom of the screen stay there as well.

My comment does get posted.

If I click the bottom Cancel, the options to away. I still need to click the top Cancel to get back to the post. If I swipe instead, I get sent back to the Posts overview page instead.

If I swipe instead of clicking either Cancel, the bottom Delete and Cancel options stay on the screen indefinitely until I click either.

nightgolfer:
![Screenshot_2023-08-08-01-26-52-78_40deb401b9ffe8e1df2f1cc5ba480b12.jpg](https://github.com/aeharding/voyager/assets/32328229/1a86ad22-1bb4-4b15-a5a9-5dd503fc3cfb)

![Screenshot_2023-08-08-01-27-05-38_40deb401b9ffe8e1df2f1cc5ba480b12.jpg](https://github.com/aeharding/voyager/assets/32328229/53789c83-fe5c-4c23-9568-4455aa2884bb)

![Screenshot_2023-08-08-01-27-13-18_40deb401b9ffe8e1df2f1cc5ba480b12.jpg](https://github.com/aeharding/voyager/assets/32328229/ffed064e-24a3-4e4f-ba62-d032cfe174e7)

![Screenshot_2023-08-07-19-34-30-25.jpg](https://github.com/aeharding/voyager/assets/32328229/8dccb555-92d7-44f6-b377-d9c7214c3c18)

nightgolfer:
Update: wrote this bug report late last night, was missing some detail. Latest version of PWA on Android.

I've now reset the app cache, and the issue seems to be gone. 

Feel free to close again, and sorry I didn't try this first 🙄

nightgolfer:
I'm just gonna mark this as completed I think. Maybe one takeaway could be the addition of a Clear Cache button in Settings? But don't know if that's possible for a PWA.

nightgolfer:
Sorry, it's still doing it. Reopening.

SrivatsanSenthilkumar:
[...]

PR #956, opened 2023-11-19
fix: make setCanDismiss synchronous

80avin:
remove workaround timeout for dismissable modals

> [!WARNING]
> Note: Issue couldn't be reproduced in a controlled manner, and was intermittent before the fix.

Fixes: https://github.com/aeharding/voyager/issues/621


src/helpers/useRefState.ts
```diff
@@ -0,0 +1,22 @@
+import { useCallback, useRef, useState } from "react";
+
+export default function useStateRef<T>(defaultValue: T) {
```
aeharding, line 3: Is this a function you made or copied from a source like SO? If so could you add the link?

src/helpers/useRefState.ts
aeharding, line 1: `useStateRef.ts`

src/helpers/useRefState.ts
80avin: This is not copied, but I made it. Just a simple abstraction so that it can be used somewhere else also.

src/helpers/useRefState.ts
80avin: Fixed.
I was using `useRefState` previously, later changed but forgot to rename file.

aeharding:
Thanks for this! :) It seems to be working quite well - the only problem I've found is on Desktop devices the confirmation page leave prompt appears when submitting a post. I pushed a fix for this, lmk if you notice any problems!

Also, there are two calls to `setTimeout(() => setIsOpen(false), 100);` in `DynamicDismissableModal.tsx`. It looks like just removing them doesn't break anything though, so I did that.

aeharding:
Thanks again!

PR #957, opened 2023-11-20
Autoselect current Lemmy instance during login

### LINKED ISSUE #782: Autoselect current Lemmy instance during login
When I choose a Lemmy instance different from the default one (`lemmy.world`), like `lemmy.ml`, it'd be nice to autoselect that version in the login flow. Right now, it just autoselects the default of `lemmy.world`:

image

image

sharunkumar:
Resolves #782

- Autoselect current Lemmy instance during login
- ux: instance selector would have the currently selected instance in the top of the list


src/features/auth/Login.tsx
```diff
@@ -61,7 +61,8 @@ export default function Login({
   const presentToast = useAppToast();
   const dispatch = useAppDispatch();
   const [servers] = useState(getCustomServers());
-  const [server, setServer] = useState(servers[0]);
```
aeharding, line 64: This doesn't work quite right. When logging in with sh.itjust.works for example, the instance is technically selected (you can press next), but there's no indication its selected in the list because its not visible.

src/features/auth/Login.tsx
```diff
@@ -61,7 +61,8 @@ export default function Login({
...
+  const { connectedInstance } = useAppSelector((state) => state.auth);
```
aeharding, line 64:
```suggestion
  const connectedInstance = useAppSelector((state) => state.auth.connectedInstance);
```

src/features/auth/Login.tsx
```diff
@@ -258,7 +259,7 @@ export default function Login({
                 onIonChange={(e) => setServer(e.target.value)}
               >
                 <IonList inset>
-                  {servers.slice(0, 3).map((server) => (
```
aeharding, line 261: The limit for up to 3 is if a self hosted instance has a lot of custom servers, so we should probably keep it.

src/features/auth/Login.tsx
sharunkumar: It was working for me, when the restriction of 3 instances is removed from the list:

image

src/features/auth/Login.tsx
sharunkumar: Wouldn't it be better in that case, since a self hosted instance user can easily select their domain when they log in?

Would the list not scroll if there are too many items on the list? I haven't checked that case :thinking:

sharunkumar:
With the recent changes I guess this is no longer relevant

PR #995, opened 2023-11-30
Add modlog link to profile action sheet

### LINKED ISSUE #979: Option to see communities mod log
**Is your feature request related to a problem?

As a user I would like to see any community's mod log

**Describe the solution you'd like**

I would like to have an option in the ellipsis pop up, that would allow me to check the modlog

**Describe alternatives you've considered**

Right now, the only way is to go to the website and getting the mod log from there

**Additional context**

It would be great to have it added here

image

0xCmdrKeen:
See title. Uses the same icon and spelling as in the admin UI, as suggested below.

Please excuse the profanity in the second screenshot. Will try to blur it out if needed.

This fixes #979.

### Screenshots
community

modlog

### DIFF
--- a/src/features/user/ProfilePageActions.tsx
+++ b/src/features/user/ProfilePageActions.tsx
@@ -1,0 +1,40 @@
+import { IonButton, IonIcon, useIonActionSheet } from "@ionic/react";
+import { ellipsisHorizontal, footstepsOutline } from "ionicons/icons";
+import { useBuildGeneralBrowseLink } from "../../helpers/routes";
+import { useOptimizedIonRouter } from "../../helpers/useOptimizedIonRouter";
+import { useAppSelector } from "../../store";
+import { usernameSelector } from "../auth/authSlice";
+import { compact } from "lodash";
+
+export default function ProfilePageActions() {
+  const router = useOptimizedIonRouter();
+  const username = useAppSelector(usernameSelector);
+  const buildGeneralBrowseLink = useBuildGeneralBrowseLink();
+  const [presentActionSheet] = useIonActionSheet();
+
+  function present() {
+    presentActionSheet({
+      cssClass: "left-align-buttons",
+      buttons: compact([
+        username && {
+          text: "Mod Log",
+          data: "modlog",
+          icon: footstepsOutline,
+          handler: () => {
+            router.push(buildGeneralBrowseLink(`/u/${username}/log`));
+          },
+        },
+        {
+          text: "Cancel",
+          role: "cancel",
+        },
+      ]),
+    });
+  }
+
+  return (
+    <IonButton fill="default" onClick={present}>
+      <IonIcon icon={ellipsisHorizontal} color="primary" />
+    </IonButton>
+  );
+}
--- a/src/pages/profile/ProfilePage.tsx
+++ b/src/pages/profile/ProfilePage.tsx
@@ -16,6 +16,7 @@ import { useContext } from "react";
 import AppContent from "../../features/shared/AppContent";
 import { PageContext } from "../../features/auth/PageContext";
 import FeedContent from "../shared/FeedContent";
+import ProfilePageActions from "../../features/user/ProfilePageActions";
 
 export default function ProfilePage() {
   const handle = useAppSelector(handleSelector);
@@ -45,6 +46,10 @@ export default function ProfilePage() {
               </IonButtons>
 
               <IonTitle>{handle}</IonTitle>
+
+              <IonButtons slot="end">
+                <ProfilePageActions />
+              </IonButtons>
             </>
           ) : (
             <>
--- a/src/pages/shared/ModlogPage.tsx
+++ b/src/pages/shared/ModlogPage.tsx
@@ -24,16 +24,21 @@ import { useBuildGeneralBrowseLink } from "../../helpers/routes";
 import { buildCommunityLink } from "../../helpers/appLinkBuilder";
 import { RootState, useAppSelector } from "../../store";
 import useCanModerate from "../../features/moderation/useCanModerate";
+import { useUserDetails } from "../../features/user/useUserDetails";
 
 export type ModlogItemType =
   GetModlogResponse[keyof GetModlogResponse] extends (infer T)[] ? T : never;
 
 export default function ModlogPage() {
-  const { community } = useParams<{ community?: string }>();
+  const { community, handle } = useParams<{
+    community?: string;
+    handle?: string;
+  }>();
 
-  if (!community) return <GlobalModlog />;
+  if (handle) return <ModlogByUserHandle handle={handle} />;
+  if (community) return <ModlogByCommunityName communityName={community} />;
 
-  return <ModlogByCommunityName communityName={community} />;
+  return <GlobalModlog />;
 }
 
 const userSelector = (state: RootState) =>
@@ -60,6 +65,18 @@ const ModlogByCommunityName = memo(function ModlogByCommunityName({
 
   if (canModerate) return <Modlog community={community.community} />;
   return <Modlog community={community.community} user={user} />;
+});
+
+const ModlogByUserHandle = memo(function ModlogByUserHandle({
+  handle,
+}: {
+  handle: string;
+}) {
+  const user = useUserDetails(handle);
+
+  if (!user) return <CenteredSpinner />;
+
+  return <Modlog user={user.user} />;
 });
 
 interface ModlogProps {
--- a/src/pages/shared/ModlogPage.tsx
+++ b/src/pages/shared/ModlogPage.tsx
@@ -22,9 +22,9 @@ import { useParams } from "react-router";
 import { getHandle } from "../../helpers/lemmy";
 import { useBuildGeneralBrowseLink } from "../../helpers/routes";
 import { buildCommunityLink } from "../../helpers/appLinkBuilder";
-import { RootState, useAppSelector } from "../../store";
+import { useAppSelector } from "../../store";
+import { userPersonSelector } from "../../features/auth/authSlice";
 import useCanModerate from "../../features/moderation/useCanModerate";
-import { useUserDetails } from "../../features/user/useUserDetails";
 
 export type ModlogItemType =
   GetModlogResponse[keyof GetModlogResponse] extends (infer T)[] ? T : never;
@@ -41,12 +41,9 @@ export default function ModlogPage() {
   return <GlobalModlog />;
 }
 
-const userSelector = (state: RootState) =>
-  state.auth.site?.my_user?.local_user_view?.person;
-
 const GlobalModlog = memo(function GlobalModlog() {
   const canModerate = useCanModerate(true);
-  const user = useAppSelector(userSelector);
+  const user = useAppSelector(userPersonSelector);
 
   if (canModerate) return <Modlog />;
   return <Modlog user={user} />;
@@ -59,7 +56,7 @@ const ModlogByCommunityName = memo(function ModlogByCommunityName({
 }) {
   const community = useFetchCommunity(communityName);
   const canModerate = useCanModerate(community?.community);
-  const user = useAppSelector(userSelector);
+  const user = useAppSelector(userPersonSelector);
 
   if (!community) return <CenteredSpinner />;
 
@@ -72,11 +69,11 @@ const ModlogByUserHandle = memo(function ModlogByUserHandle({
 }: {
   handle: string;
 }) {
-  const user = useUserDetails(handle);
+  const user = useAppSelector((state) => state.user.userByHandle[handle]);
 
   if (!user) return <CenteredSpinner />;
 
-  return <Modlog user={user.user} />;
+  return <Modlog user={user} />;
 });
 
 interface ModlogProps {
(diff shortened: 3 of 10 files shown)


sallyFoster:
@0xCmdrKeen The icon that is currently used to see mod logs as a moderator account is the following. Maybe you can use this one too?

image

0xCmdrKeen:
@sallyFoster sure can do.

aeharding:
Thanks for the PR!

The mod log was intentionally not added to Voyager for regular users because it contains a lot of abusive content that, by nature, cannot have entries removed or filtered, which may be a problem for app stores.

So, that is why I cannot merge this at the moment. In the future we might be able to add it, with a warning, buried further in the app where someone might not accidentally press it.

sallyFoster:
@aeharding 

I understand. In the future, is it possible that you guys can keep the style of this PR, and have a toggle to activate it with a warning deep in settings? It might be a hassle for users to have to click a million windows each time to see the mod log of each particular community.

aeharding:
Yeah. I was thinking perhaps put it at the bottom of the sidebar, perhaps. Then its just one extra click but also much less likely to be accidentally tapped. But I'll do some more research on whats OK for the app stores first.

0xCmdrKeen:
@aeharding okay, that makes sense. I certainly don't want you to get kicked off the App Store.

How about this: instead of showing normal users the entire modlog for a certain community, we only show them the modlog for their own user? This would likely require a little more work since it doesn't look like that's implemented yet, but the Lemmy API DOES support filtering the modlog by a specific user. 

This would not run the risk of ever exposing the user to any unexpected profanity or NSFW content, but it would still be useful, since Lemmy doesn't give you any feedback when moderator actions were taken on your own content, and few mods seem to make the effort to send a PM to let you know about this either. A setting to remove that filter could still be added later, once you figure out a good way to do this.

aeharding:
> @aeharding okay, that makes sense. I certainly don't want you to get kicked off the App Store.
> 
> How about this: instead of showing normal users the entire modlog for a certain community, we only show them the modlog for their own user? This would likely require a little more work since it doesn't look like that's implemented yet, but the Lemmy API DOES support filtering the modlog by a specific user.
> 
> This would not run the risk of ever exposing the user to any unexpected profanity or NSFW content, but it would still be useful, since Lemmy doesn't give you any feedback when moderator actions were taken on your own content, and few mods seem to make the effort to send a PM to let you know about this either. A setting to remove that filter could still be added later, once you figure out a good way to do this.

That's a great idea!

0xCmdrKeen:
> That's a great idea!

Alright, I will try to take a stab at it. Will update this PR when and if I get it working.

aeharding:
Sounds good. Another option is instead of per-community we could show all mod actions against your account. We'd have to put it under the profile page or something if we went that route. What do you think of that?

sallyFoster:
@0xCmdrKeen @aeharding Would it be too much work if you implemented it for both specific communities and all communities 🥺

0xCmdrKeen:
@sallyFoster no that's a good point, I was just thinking that there should be a way to do that, and then I found this screen and thought that might a good place to put that:

image

0xCmdrKeen:
Alright, so I added a mod log link to the instance "more actions" sheet, and I made so that only moderators can see the whole thing. A regular user will only see items involving themselves. 

When using the app in a logged-out state (i.e. as a guest user), the mod log links do not show up at all, since they otherwise *would* end up showing the whole log.

A small issue I've noticed is that when viewing the whole instance log, community bans do not show the name of the community the user was banned from (see first screenshot). This isn't really a problem when viewing a community modlog, since the name of the community is shown in the header, but when viewing a server modlog, it's not very useful, so I'm trying to figure out a good way to add that information here (see second screenshot). 

Please LMK if you have any suggestions.

image
image

sallyFoster:
@0xCmdrKeen 

So currently for moderator accounts there is a button that allows to perform moderation actions including checking all communities modlog. Maybe you can check for wether a username is moderator or not and repurpose this button to only check for modlog if the user is not a moderator? 

image

0xCmdrKeen:
@sallyFoster I'm not sure what you're trying to say here. I believe that button already DOES check if you're a moderator, otherwise it won't show up at all.

Are you trying to say that if a user *isn't* a moderator, it should just show a link to the mod log instead of the whole moderator menu? If so, that's certainly possible but I'm not sure if it's necessary to make it more visible. But I'll let @aeharding make the call on that.

sallyFoster:
@0xCmdrKeen Yeah, I was referring to always having the button visible, and changing the button text to say “Mod Log” when the user is not a moderator since having a moderation menu is not necessary for non-moderators. What I mean is that we can just have the button say “Mod log” and open the mod log on tap for non-moderator users.

0xCmdrKeen:
@sallyFoster ah, okay, thanks for clarifying. Like I said, I'm not convinced it needs to be that prominent in the app and I DO want that feature too. But I don't think that I should be making that call, so I'll wait for Alex's response on this.

aeharding:
Yeah, I don't think it should be that prominent in the profile.

But I do think the profile page is a good place to put it. What if we added an ellipse in the top right of the header to open an action sheet, with the only action being "Mod Log" or something?

This would also allow everyone to see mod actions against their account, not just regular users.

0xCmdrKeen:
@aeharding that's a pretty good idea, actually. In that case, I wonder if it would make sense to only show the other mod log links for actual mods, since the one on the instance page would show the same thing for a regular user, and the one on a community page would show a subset of that.

As much as I would like to see this feature as a regular user, I don't think it needs to be plastered all over the app.

aeharding:
OK, so for this PR, add the ellipsis with button on the profile header but not add other places sound good?

0xCmdrKeen:
Can do, but what about showing the other links for moderators only? 

I.e. community mod log if the user has permission to moderate that community, and instance mod log link if they have permission to moderate on that instance.

sallyFoster:
@0xCmdrKeen But doesn't the Moderator Zone button already do for moderators what you are proposing?

0xCmdrKeen:
@sallyFoster you mean the ellipsis on the user profile that Alex suggested?

Because I though putting the modlog links in the community menus was the whole point of this issue...

But I think it's certainly also possible to make the former only show up for regular users, and the latter only for mods.

aeharding:
Mods already can view the mod log from the mod zone or from the community page.

This PR should just address viewing logs for normal users, I thought we agreed to just allow the user to view moderation actions against their account (e.g. through their profile) only.

0xCmdrKeen:
Okay, I can remove the other two. Should the link for normal users only show up if they can't moderate then?

aeharding:
I think it makes sense to show regardless, because mods may also want to see actions against their account.

0xCmdrKeen:
Okay, the mod log buttons from the community and instance more action sheets are now gone, instead, the profile page has an ellipsis button with a "more actions" menu that shows a single entry to view the user's mod log.

I also fixed the community ban message to include the name of the community, and included the user name in the page title for the user-specific mod log.

0xCmdrKeen:
@sallyFoster the message should show the community name only for local communities (i.e. on the same server) and the full name for remote ones.

And no, I removed both of the other links, I thought that's what everyone had agreed on.

src/pages/shared/ModlogPage.tsx
```diff
@@ -44,13 +58,34 @@ const ModlogByCommunityName = memo(function ModlogByCommunityName({
...
+  return <Modlog community={community.community} user={user} />;
 });
 
-function ModlogByCommunity({ community }: { community?: Community }) {
+const ModlogByUserHandle = memo(function ModlogByUserHandle({
+  handle,
+}: {
+  handle: string;
+}) {
+  const user = useUserDetails(handle);
```
aeharding, line 75: You could probably just make this a selector, there's a lot of unecessary stuff in useUserDetails.

src/pages/shared/ModlogPage.tsx
```diff
@@ -22,20 +22,34 @@ import { useParams } from "react-router";
...
+  return <GlobalModlog />;
 }
 
+const userSelector = (state: RootState) =>
+  state.auth.site?.my_user?.local_user_view?.person;
+
 const GlobalModlog = memo(function GlobalModlog() {
-  return <ModlogByCommunity />;
+  const canModerate = useCanModerate(true);
+  const user = useAppSelector(userSelector);
```
aeharding, line 49: inline the selector function here

src/pages/shared/ModlogPage.tsx
```diff
@@ -70,13 +107,19 @@ function ModlogByCommunity({ community }: { community?: Community }) {
...
     },
-    [client, community],
+    [client, community, user, mod],
   );
 
   const renderItemContent = useCallback((item: ModlogItemType) => {
     return <ModlogItem item={item} />;
   }, []);
 
+  const title = community
```
aeharding, line 117: Refactor to a function:

```ts
const title = (() => {
  if (community) return getHandle(community);
  if (user) return getHandle(user);

  return "Mod"
})()

**Comment by sallyFoster** [comment]
@0xCmdrKeen I am very confused. Can you please share screenshots with your changes?

I think that what @aeharding was saying is that we should show the mod log of the particular user and only of that particular user in any community they want to check it on, For example:


image

And the mod log of that same particular user in all communities, which is the one that you added with the ellipsis. 

@aeharding just doesn't want the non-mod user to see any mod log that is not theirs. At least that is what I understood.

**Review by aeharding — COMMENTED** [review]
Looks great, few things!

File: src/pages/shared/ModlogPage.tsx (hunk 2 above)
Comment (0xCmdrKeen) [reply src/pages/shared/ModlogPage.tsx:49]: The same selector function is used again below, on line 62

File: src/pages/shared/ModlogPage.tsx (hunk 1 above)
Comment (0xCmdrKeen) [reply src/pages/shared/ModlogPage.tsx:75]: It looks like the only thing I need from that hook is this:

```ts
const userByHandle = useAppSelector((state) => state.user.userByHandle);
const user = userByHandle[handle];
```

Can I combine the two into a single line like this:

```ts
const user = useAppSelector((state) => state.user.userByHandle[handle]);
```

...or is that liable to break Redux somehow? I just tested it and it DOES seem to work.

src/pages/shared/ModlogPage.tsx
0xCmdrKeen: Ok, can do.

src/pages/shared/ModlogPage.tsx
0xCmdrKeen: Sure can!

src/features/user/ProfilePageActions.tsx
aeharding, line 1: Rename to `MyUserPageActions` (there's already `UserPageActions` for not-myself actions)

src/pages/profile/ProfilePage.tsx
```diff
@@ -45,6 +46,10 @@ export default function ProfilePage() {
               </IonButtons>
 
               <IonTitle>{handle}</IonTitle>
+
+              <IonButtons slot="end">
+                <ProfilePageActions />
```
aeharding, line 51: You will also need to add this to `<ProfilePage>` (if not `isSelf`).

ProfilePage = `/profile`
UserPage = `/profile/lemmy.world/u/aeharding` (for example) - you can also access your user data here

src/features/user/ProfilePageActions.tsx
0xCmdrKeen: Right, that's why I named it `ProfilePageActions` in the first place ;)
I figured that would make the distinction clear, since the corresponding views are named `ProfilePage` and `UserPage`.

src/pages/profile/ProfilePage.tsx
0xCmdrKeen: I think this already works as expected, or at least it works in a way that makes sense:

Go to `/profile` -> "more actions" menu shows link to user modlog
Go to `/profile/[instance]/u/[my_username]` -> no "more actions" menu 
Go to `/profile/[instance]/u/[other_username]` -> "more actions" menu shows links to send message and block user

Unless you also want the modlog link to show in the second scenario. But I don't think that's necessary, since it can already be accessed from the profile.

aeharding: Yes, the ellipse should always show if you're viewing your own profile. To the user their the same page. The code should probably not be duplicated here but thats beyond the scope of this ticket.

0xCmdrKeen: Okay, gotcha. I'll get it done.

aeharding:
Thanks!

s38b35M5:
Great addition to the app!

Is it intended to not include moderation reason in this modlog? I can currently only see: action, post/comment description, mod/admin name.

 I still have to use the web interface to see moderation reason.

0xCmdrKeen:
@s38b35M5 thanks, and that's a good point. Since this PR is already closed, this should probably be a separate issue. I'll just go ahead and create one.

sallyFoster:
@0xCmdrKeen For some reason, for me, the community where the post was deleted from and the mod who deleted the post are not showing. The only things that are showing are the post title and the action taken.

0xCmdrKeen:
@sallyFoster can you add this comment to #1046? I think the entire modlog display probably needs a bit of TLC.

Also, the PR didn't break this, it just exposed it.

sallyFoster:
Added

Issue #998, opened 2023-12-01
Allow customizing the list of default instances


0xCmdrKeen:
**Is your feature request related to a problem? Please describe.**
When logging in to an account, Voyager currently shows a list of only 3 default instances, with the option to manually type the URL/domain for a custom instance you'd like to use on a per-case basis (see first screenshot).

**Describe the solution you'd like**
What I'd like to see is an option in the settings to customize the list of default instances, so you can permanently add your preferred instances to the list. Lunar currently does this very elegantly (see second screenshot). 

In Lunar, you can access this screen from the top level settings menu via an entry called "Manage Instances", but it could certainly also be placed within one of the existing categories in Voyager's settings menu (e.g. "General") if that seems preferable. The advantage of that would mainly be not having to come up with an appropriate icon choice.

Basically, the entire list is editable (meaning you can remove any of the default instances if you'd like), but you can always get the default list back using the "Reset Instances List" button. It probably isn't necessary for the list to also support re-ordering by dragging and dropping if that means a lot of extra work (it's not orderable in Lunar either), since one can always just remove and re-add instances in the order desired. 

**Describe alternatives you've considered**
Not giving users a simple and accessible way to add custom instances will likely contribute to increased centralization in the fediverse by driving more traffic to the "big 3", since manually typing a custom instance name gets old rather quick, and people are naturally lazy.

**Additional context**

<img width="350" alt="Voyager" src="https://github.com/aeharding/voyager/assets/98132670/4013902d-ff3b-4bf0-a2ae-46a156bf97e2">

<img width="350" alt="Lunar" src="https://github.com/aeharding/voyager/assets/98132670/e29f2c78-16a1-4316-b311-6cc3d8ea7aa1">

0xCmdrKeen:
I did some work on this and I was able to code a "Manage instances" page that looks and functions exactly like that from Lunar (second screenshot). Newly added servers are validated (similarly to the process used in the `Login` component) before they are added to the list to prevent users from adding random junk.

Happy to contribute a PR, however, it currently only keeps the server list in local state. Would need some advice on how to best wire it into the app settings to make this persistable. It doesn't look like the ability to store a custom server list currently exists anywhere in the app.

80avin:
@0xCmdrKeen  You may want to use [db](https://github.com/aeharding/voyager/blob/13672ab82eaaff84866985c7ae49cc0083d71d34/src/services/db.ts#L617) or localStorage ?

[...]

PR #1010, opened 2023-12-03
feat: Manage instances list

### LINKED ISSUE #116: Custom server viewing when anonymous

0xCmdrKeen:
Took a stab at #998. 

This adds a "Manage instances" screen to the Settings menu which allows editing the default list of Lemmy instances used by the app. Instances can be added and deleted, reordering is currently not implemented. There is a button to restore the default list in case the user messes up.

Newly added instances are verified to be Lemmy servers (using the `getSite()` call) before being added to the list. The list is stored in the database to make it persistent.

The Login and LoggedOut screens will now use the custom server list (this fixes #116). When choosing the "Other" option on the Login screen, if the login was successful, the instance will be added to the list automatically. 

As I mentioned in the issue discussion, I'm not entirely decided whether the "Manage instances" setting should be a top-level item or go into one of the existing categories, but of course this can be changed before merging. 

I'm also happy to change the particulars about where in the Redux state and the database the list is stored.

### DIFF
--- a/src/TabbedRoutes.tsx
+++ b/src/TabbedRoutes.tsx
@@ -27,6 +27,7 @@ import InboxAuthRequired from "./pages/inbox/InboxAuthRequired";
 import UpdateAppPage from "./pages/settings/UpdateAppPage";
 import AppearancePage from "./pages/settings/AppearancePage";
 import CommunitySidebarPage from "./pages/shared/CommunitySidebarPage";
+import ManageInstancesPage from "./pages/settings/ManageInstancesPage";
 import RedditMigratePage from "./pages/settings/RedditDataMigratePage";
 import ProfilePage from "./pages/profile/ProfilePage";
 import ProfileFeedHiddenPostsPage from "./pages/profile/ProfileFeedHiddenPostsPage";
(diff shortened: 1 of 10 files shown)


0xCmdrKeen:
Just noticed the part in the project README about self-hosting and using a custom server list, and finally understood the remaining code in `services/app.ts` that I couldn't quite figure out before.

Obviously, this feature will likely not be desirable in that scenario. Will figure out a way to disable it. And easy solution would be to only have it enabled in the native app, since that doesn't use an external server list anyways, but it would also prevent it from showing up in the development preview version or on general purpose deployments like vger.app. But I'll think of something.

aeharding:
Thanks for the PR. This one is a bit involved just because the custom servers logic has a lot of edge cases depending on the users' platform, self-hosted config etc so that is why I haven't gotten to reviewing yet.

0xCmdrKeen:
@aeharding agreed, and as you may have noticed, I already found some of those edge cases myself :)

But yes, this is certainly a more complex feature and I understand that it requires a thorough review to make sure it doesn't unintentionally break anything.

aeharding:
Hi @0xCmdrKeen thanks again for you PR and effort on this!

I think this PR fits better with Voyager's old login flow. However this is definitely something I would like to add, in some form or another.

With Voyager's new log in flow being an autocomplete list, I think it would be cool if instances could be remembered for next time. Essentially a "recently logged in" list, or something of that nature.

That's a bit of an easier user experience, and would be helpful for people that don't want to take the time to configure a list of custom instances in Settings.

So that's why I'm closing this PR. But again I do want to address #998.

I'm hoping to get a PR out this weekend that will allow browsing and switching between anonymous (logged out) instances. Once that gets merged, I'd be happy to accept a PR with some form of autocomplete/remembering recently logged in instances. We can discuss more on #998, if you'd like.

PR #1020, opened 2023-12-05
Update HidingSettings.tsx to remove the double occurance of inside in the helper text of Marking Read/Hiding settings

SrivatsanSenthilkumar:
I know its a trivial commit but @aeharding you offered us the oppurtunity to do it so here it is.


SrivatsanSenthilkumar:
[Me contributing to open source software by adding one line in a txt file](https://i.kym-cdn.com/photos/images/newsfeed/001/584/180/0c2.jpg)

PR #1021, opened 2023-12-05
Add pride icon

fer0n:


SrivatsanSenthilkumar:
Maybe you can use the psychedelic icon's background. That i more colouful and, for a lack of a better word, rainbow-y

fer0n:
@SrivatsanSenthilkumar here's a few (lower effort) alternatives:
psy
psy 2
Voyager icon

These alternatives would certainly look a slightly better with some tweaking, but especially the rainbow-on-rainbow is a bit too much imo and it's hard to tell what's going on when looking at it in icon size on a phone. I like how the flag is a more diegetic way of bringing that theme into the icon and it pops more with a different background. I also tried using the flag colors and working them into the background somehow, but most of them don't go together well outside the flag. I'm reasonably happy with the current version.

That said I'm sure there can be better icons than this, with a similar or an entirely different style. Anyone who wants to is free to submit something. I've also attached the affinity designer file in case you want to take a look yourself or play around with it.

[Voyager icon.zip](https://github.com/aeharding/voyager/files/13556878/Voyager.icon.zip)

fer0n:
@aeharding should be fixed now

SrivatsanSenthilkumar:
> @SrivatsanSenthilkumar here's a few (lower effort) alternatives:
> 
> 
> 
> 
> For reference:
> 
> 
> These alternatives would certainly look a slightly better with some tweaking, but especially the rainbow-on-rainbow is a bit too much imo and it's hard to tell what's going on when looking at it in icon size on a phone. I like how the flag is a more diegetic way of bringing that theme into the icon and it pops more with a different background. I also tried using the flag colors and working them into the background somehow, but most of them don't go together well outside the flag. I'm reasonably happy with the current version.
> 
> That said I'm sure there can be better icons than this, with a similar or an entirely different style. Anyone who wants to is free to submit something. I've also attached the affinity designer file in case you want to take a look yourself or play around with it.
> 
> [Voyager icon.zip](https://github.com/aeharding/voyager/files/13556878/Voyager.icon.zip)

I have to agree the original design you presented is not over top and yet has a clear reference to pride month. The other do seem like a overload of colours. Anyway thanks for indulging my requests.

PR #715, opened 2023-08-28
feat: Account Switcher on Long Pressing profile icon in Navbar

rahulakrishna:
- open account switcher modal on long pressing profile icon in navbar
- if in comments view, resolve the post in new instance using the `ap_id` and replace the route without animation
- if in community view, replace the `actor` in url with the new instance name
- add hammerjs

Demo: 

https://github.com/aeharding/voyager/assets/10240002/fc8263b2-2556-4012-bbee-a206cc498f13

Resolves: #85

PR #1031, opened 2023-12-08
feat: Quick account switcher

0xCmdrKeen:
Since #715 has blocked and inactive for a while, and I really miss this feature from Apollo, I thought I'd try my own version of this.

This makes tapping the Profile button in the tab bar pull up the account switcher IF the profile page is already active in the profile tab (and hasn't been scrolled down on).

This means tapping the Profile button TWICE will bring up the account switcher also.

If you open the Profile page by tapping the button once, and then navigate away by clicking one of the links on it, tapping the profile button again will bring you back to the profile page, as before.

https://github.com/aeharding/voyager/assets/98132670/1f41e4db-5917-44bb-ac38-1c94948861b5

### DIFF
--- a/src/TabBar.tsx
+++ b/src/TabBar.tsx
@@ -139,7 +139,7 @@ const TabBar: CustomTabBarType = forwardRef(function TabBar(props, ref) {
     if (scrollUpIfNeeded(activePageRef?.current)) return;
 
     // if the profile page is already open, show the account switcher
-    if (router.getRouteInfo()?.pathname === "/profile")
+    if (location.pathname === "/profile")
       presentAccountSwitcher({ cssClass: "small" });
 
     router.push("/profile", "back");
--- a/src/TabBar.tsx
+++ b/src/TabBar.tsx
@@ -12,7 +12,6 @@ import {
   IonLabel,
   IonTabBar,
   IonTabButton,
-  useIonModal,
 } from "@ionic/react";
 import { totalUnreadSelector } from "./features/inbox/inboxSlice";
 import useShouldInstall from "./features/pwa/useShouldInstall";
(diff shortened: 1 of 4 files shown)


src/features/auth/PageContext.tsx
```diff
@@ -59,6 +62,13 @@ interface IPageContext {
     comment?: CommentView,
     comments?: CommentView[],
   ) => void;
+
+  presentAccountSwitcher: (
```
aeharding, line 66: Let's have this function take no arguments, and then add a function in PageContext to call presentAccountSwitcher with cssClass = small.

src/TabBar.tsx
```diff
@@ -136,6 +138,10 @@ const TabBar: CustomTabBarType = forwardRef(function TabBar(props, ref) {
 
     if (scrollUpIfNeeded(activePageRef?.current)) return;
 
+    // if the profile page is already open, show the account switcher
+    if (router.getRouteInfo()?.pathname === "/profile")
```
aeharding, line 142: `location.pathname` instead to be consistent with the rest of the component :)

aeharding (commented):
Looks good, I like your idea!

src/features/auth/PageContext.tsx
0xCmdrKeen: Sounds good

src/TabBar.tsx
0xCmdrKeen: No problem

aeharding:
Looks good! I had one problem where the presentingElement of the login modal from the account switcher wasn't properly attaching so I pushed a fix for that.

PR #1041, opened 2023-12-09
Add support for upvoted/downvoted

aeharding:


aeharding:
I'm going to mark this blocked until lemmy.world upgrades to lemmy 0.19 and I can test there. (There's some performance issues with these requests.)

PR #1058, opened 2023-12-11
Add support for inline code in inline markdown

### LINKED ISSUE #1051: Inline code in post titles doesn't render
**Describe the bug**
Apparently,  Lemmy allows Markdown to be used in post titles ([example](https://programming.dev/post/6973774)). This renders properly on the web client, but in Voyager, the Markdown part is simply ignored, making the title read funny.

**To Reproduce**

1. Go to search tab
2. Paste this post URL: https://programming.dev/post/6973774
3. Tap on "Visit post"

**Expected behavior**
Markdown should be rendered properly, not simply ignored.

**Screenshots**
image

**Desktop (please complete the following information):**

- OS: Windows 11
- Browser: Brave 1.61.101
- Voyager 1.30.2

**Smartphone (please complete the following information):**

- Device: iPhone XS
- OS: iOS 16.6.1
- Voyager 1.30.2

**Additional context**
I'm not sure WHAT subset of Markdown exactly is allowed here, but inline code clearly is. It appears that there already is a `InlineMarkdown` component to handle this exact scenario, but the current configuration does not appear to allow inline code.

I tried adding `"code"` to the `allowedElements` and `code: "code"` to `components` and it seems to solve the issue reasonably well (at least in the browser, can't test the native app as I don't have the toolchain set up for that). Can send a PR but not sure if it's worth the effort for just those 2 lines.

0xCmdrKeen:
This fixes #1051

### DIFF
--- a/src/features/shared/InlineMarkdown.tsx
+++ b/src/features/shared/InlineMarkdown.tsx
@@ -24,6 +24,7 @@ export default function InlineMarkdown({ children }: InlineMarkdownProps) {
           "h3",
           "h4",
           "h5",
+          "code",
         ]}
         components={{
           a: "span",
@@ -39,6 +40,7 @@ export default function InlineMarkdown({ children }: InlineMarkdownProps) {
           h3: "span",
           h4: "span",
           h5: "span",
+          code: "code",
         }}
       >
         {children}


aeharding:
Cheers!

PR #1087, opened 2023-12-15
Change post/comment editor abort action title to "Discard"

### LINKED ISSUE #1059: When editing a comment with unsaved changes, pressing "Cancel" asks to confirm deletion
**Describe the bug**
See title. When editing a comment and some changes have been made, pressing the "Cancel" button or swiping down shows a confirmation message asking to confirm deletion. This makes sense when canceling a new comment in progress, but it is somewhat confusing when editing a comment since the user might think their comment will end up getting deleted if they go ahead. It's also red which implies a potentially destructive action, thus heightening the sense of danger.

It would be less confusing if the button said "Discard" instead. This would in fact be clear and unambiguous in all situations (new comment or editing comment), so it would not need any special case handling.

**To Reproduce**

1. Go to Profile page
2. Click on ellipsis button next to a comment
3. Select "Edit"
4. Type some text
5. Press cancel (or swipe down)

**Expected behavior**
Instead of "Delete", the confirmation button should say "Discard".

**Screenshots**
screenshot

**Additional context**
For even more clarity, the button could say "Discard comment" if it's a new comment and "Discard changes" when editing.

0xCmdrKeen:
When pressing "cancel" on a post or comment that has unsaved changes, the button to confirm deletion now says "Discard" instead of "Delete".

image

This fixes #1059.

### DIFF
--- a/src/features/shared/DynamicDismissableModal.tsx
+++ b/src/features/shared/DynamicDismissableModal.tsx
@@ -51,7 +51,7 @@ export function DynamicDismissableModal({
 
     await presentActionSheet([
       {
-        text: "Delete",
+        text: "Discard",
         role: "destructive",
         handler: () => {
           clearRecoveredText();


aeharding:
Thanks!

PR #1096, opened 2023-12-17
Feature: dialog for link input

0xCmdrKeen:
I didn't create an issue for this, but Apollo used to show an input dialog for inserting links into Markdown text (i.e comments or posts). This is both helpful for people who are not super familiar with the syntax, and also for not having to fiddle with text selection on mobile, which can be annoying.

This PR adds such a feature to Voyager (see screenshot). Both fields are optional, if left empty, it will behave exactly as before (i.e. insert `[](url)` and place the cursor between the brackets). If only the link text is given but no URL, `[text](url)` will be inserted, with the "url" part preselected. If both are given, the cursor will be placed at the end.

Long pressing the URL button will skip the dialog and directly invoke the old behavior.

Apollo used to show a link to a random charity as the URL placeholder, but I didn't want to do that without asking first. Can definitely add this if desired (please provide a list of acceptable charities in this case).

Another convenience feature I loved was automatically pre-filling the URL input if a link was detected in the clipboard, but with iOS now requiring permission for access each time, I'm not sure it wouldn't be more annoying than useful. There IS [a capacitor package](https://capacitorjs.com/docs/apis/clipboard) for interacting with the clipboard, but it doesn't say anything regarding permissions.

screenshot


sharunkumar (approved):

sharunkumar:
Quite a lot of conflicts compared to the current version. Should be easy to re-implement the changes after rebasing though

aeharding:
Hi, thanks for the PR!

I just got around to reviewing. Nice work, I like the auto highlighting functionality!

In fixing the conflict I made a few changes:

- Fixed the conflicts
- Removed the long press functionality. The `md-link` was doing some fancy stuff where if you highlight something it would use that highlighted text. We can re-add if you'd like, but this wasn't implemented in the PR
- Added support for autofilling the alert if you have selected text on screen
- Added support for auto focusing relevant text box
- Added type=url to the URL text box for better keyboard support

When you have a moment, please let me know if this all looks good to you and we can get this in. Thanks!

aeharding:
I'm going to get this merged for the next release. If you have any concern please open an issue/PR! Thank you.

PR #1141, opened 2024-01-05
Feature: remote content visibility detection

0xCmdrKeen:
## Summary

Since Lemmy 0.19 still seems to have outgoing federation issues, I thought it might be helpful if Voyager could check whether my posts or comments have propagated to other instances yet, and display a notification when they haven't.

This PR implements that feature. Any item posted by the currently active user that isn't visible on the home instance of the community it was posted to will have a tiny yellow warning symbol next to it (see screenshots). Only the current user's own content is checked, and only if it wasn't posted to a local community. Tapping the icon will open a dialog explaining its meaning, and gives the user an opportunity to do a manual re-check, since lookups are cached to avoid excessive network traffic.

Currently, the feature is always on and automatically works in the background, but I can add an option in the settings to turn it off completely, or make it manual (via action sheet menu item). I could also add automatic re-checking of non-federated items based on a certain time interval (and make it configurable if desired).

## Screenshots

screenshot

screenshot

### DIFF
--- a/src/features/comment/Comment.tsx
+++ b/src/features/comment/Comment.tsx
@@ -22,6 +22,7 @@ import ModeratableItem from "../moderation/ModeratableItem";
 import useCanModerate from "../moderation/useCanModerate";
 import ModqueueItemActions from "../moderation/ModqueueItemActions";
 import { ActionsContainer } from "../post/inFeed/compact/CompactPost";
+import Visible from "../labels/Visible";
 
 const rainbowColors = [
   "#FF0000", // Red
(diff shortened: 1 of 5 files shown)


0xCmdrKeen:
Note: it's still making duplicate requests for some reason, I'm putting it into draft mode until I can figure it out.

aeharding:
Hi @0xCmdrKeen, thanks for the PR! This is interesting, but I don't think its something I feel comfortable adding directly to Voyager. Lemmy devs are working on improving federation so that instances should not lag on federation.

0xCmdrKeen:
@aeharding yes, it appears that the federation issues with Lemmy 0.3 that this PR was intended to solve haven't been occurring anymore for a while now, so I think I'll just close this for now.

PR #1158, opened 2024-01-14
Add sign up, refactor login flow

aeharding:
This is quite similar to Mastodon's approach to sign up and log in.

TODO

- [x] add filters for join server list
- [x] Default server support


aeharding:
Created https://github.com/tgxn/lemmy-explorer/issues/180 to track adding a registration mode filter for those who want to sign up instantly :)

Issue #1154, opened 2024-01-11
Remove upvote/downvote buttons in post menu


slrgt:
**Is your feature request related to a problem? Please describe.**
Important buttons like crosspost and report are hidden and require a scroll because there are too many buttons in the menu

**Describe the solution you'd like**
It would be convenient if these buttons were displayed without needing to scroll and the upvote/downvote buttons are redundant since they're already present next to the menu button.

This is how it could look by default:

![image](https://github.com/aeharding/voyager/assets/114287507/df4120e2-624b-46e7-9545-5d2db8af8724)

aeharding:
Thanks, however this is a duplicate of https://github.com/aeharding/voyager/issues/417

PR #1167, opened 2024-01-18
Removes upvote and downvote from MoreActions.tsx

bartivs:
As described in #1154


aeharding:
Hi, thanks for the PR, however this was already rejected, see reason here:

https://github.com/aeharding/voyager/issues/417

PR #1214, opened 2024-02-04
Add option to collapse comments recursively by default

### LINKED ISSUE #1153: Add option to collapse comments recursively
It would be nice to have an option to, if "Collapse Comment Threads" is enabled, to further collapse all child comments as well by default. 

If enabled, when you open a post it should only show Level 0 (i.e. parent) comments. When tapping a comment to expand its children it should only show Level 1 comments. Tapping on a Level 1 comment should show Level 2 comments, etc.

This would allow easier navigation of highly commented posts, with more of a "breadth first" experience rather than "depth first." I believe Baconreader had this behavior.

ShadowAndFlame:
Resolves #1153 and adds an option to cause comments and all children to start collapsed.

image

image

Determining which mode is selected to fetch the correct depth of comments when tapping "n more replies" turned out to be tricky, and I ended up creating a new app selector rather than modifying defaultCommentDepthSelector. Maybe not the most elegant solution but I didn't want to risk messing with other comment behavior.

Since the "Always" setting was changed to "Root Only", a database migration was required.

Please let me know how I did. This is my first contribution on GitHub and my first experience with TypeScript.


aeharding:
Hi! Thank you for the PR! It looks like there are some lint issues preventing successful build. I'll try to review this as soon as I have time.

ShadowAndFlame:
Thanks, I moved it to draft when I saw it failed those checks. I've been away but I'll take a look when I can.

aeharding (approved):
Working great! Thank you!

PR #1294, opened 2024-03-08
Replace supersub markdown plugin with custom for lemmy

aeharding:
This fixes some edge cases in super sub parsing


aeharding:
Good test case:

https://sh.itjust.works/comment/9457351

```
For those who dont understand the equation, E = Energy, specifically Kenetic Energy in this case, p = momentum, which is determined by mass and velocity, and m = mass. You wanna know something else? a^2 + b^2 = c^2 is the same fucking thing as E^2 = p^2 + m^2. 
```

Issue #1331, opened 2024-03-15
Display hostname when viewing posts in compact mode


bishtawi:
**Is your feature request related to a problem? Please describe.**
When viewing posts in compact mode, I would like to see the domain name of the link I am about to click on. I dont want to click on a link without first knowing where it leads to.

**Describe the solution you'd like**
When in compact mode, it would be nice if posts also displayed the domain name of the link. I understand its compact mode so I don't expect to see the full url but I think at least showing the host name is a good idea.

**Describe alternatives you've considered**
Currently I click into the comments so I can see the link. Its cumbersome as it adds an extra step. Often times I want to read the article before diving into the comments.

**Additional context**
Its good security practice to at least know the site you are about to visit

PR #1348, opened 2024-03-20
Display domain on compact posts

bishtawi:
Addresses https://github.com/aeharding/voyager/issues/1331

I love your app! Thought I would give it a go and add this feature since it seemed straight forward.

Screenshot:
Screenshot from 2024-03-20 00-32-41

CSS can probably use some work. Not sure how much effort it is, but including the site's favicon to the left of the domain name could be a nice touch.

### DIFF
--- a/src/features/post/inFeed/compact/CompactPost.tsx
+++ b/src/features/post/inFeed/compact/CompactPost.tsx
@@ -138,10 +138,6 @@ export default function CompactPost({ post }: PostProps) {
     (state) => state.settings.appearance.compact.thumbnailsPosition,
   );
 
-  const compactShowDomainNames = useAppSelector(
-    (state) => state.settings.appearance.compact.showDomainNames,
-  );
-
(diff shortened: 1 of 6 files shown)


aeharding:
Hi, thanks for the PR!

We probably should add an option to configure this in Settings since I don't think everyone will want this in their feed. Can you look at that? There is already a section for compact appearance settings :)

aeharding:
@bishtawi that's actually a really good point, and since we don't allow disabling the URL display anywhere else in the app its probably best to be consistent for now and force the URL to always show. I'll revert that commit. Thanks for pointing that out.

BTW nice work on extracting out the `parseUrlForDisplay` function! I just tried out the PR in more detail made some tweaks to the CSS/display of URL:

 1. URL now displays in parenthesis
 2. URL has max length, so long URLs (or large font sizes) don't break the layout
 3. URL for images are hidden
 4. URL now displays in title of post

Let me know what you think!

PR #1356, opened 2024-03-23
Ignore Typescript Incremental Cache File

sharunkumar:
When running `tsc --incremental` this file is generated. This can be ignored in git/docker


aeharding:
Thanks!

PR #1371, opened 2024-03-26
(chore) vger.app s3 bucket auto upload

aeharding:


.github/workflows/cd.yml
```diff
@@ -1,11 +1,48 @@
...
+          args: --acl public-read --follow-symlinks --delete
+        env:
+          SOURCE_DIR: dist
+          AWS_S3_BUCKET: ${{ secrets.AWS_S3_BUCKET }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+
+      - name: Compress artifacts
+        run: |
+          zip -r Voyager-Web-${{ github.ref_name }}-Build.zip dist
```
sharunkumar, line 38: `server.mjs` not included in zip file?

aeharding: No, this is just the static bundle. `server.mjs` is not needed and actually will be removed soon :)

vger.app now runs in a plain s3 bucket.

PR #1379, opened 2024-03-29
Change runtime docker from node to nginx

aeharding:
Since upgrading to require Lemmy 0.19, Voyager no longer requires a backend.

So to simplify the runtime, I've switched out Node to nginx. There is a small script `generate_config.sh` that generates the server JSON list from `CUSTOM_LEMMY_SERVERS` on startup.


aeharding:
Requesting the review of those that have worked on the Voyager docker image- @sharunkumar, @mbentley, @jonesetc, @Crazybus (sorry for the spam if you are not interested!)

sharunkumar:
Looks like the port has been changed to 80 as default, might break existing configurations that forward to port 5314

Readme can be updated to show this change: `docker run --init --rm -it -p 5314:80 voyager`.

I built and ran the container locally with the above command and seems to be working fine

aeharding:
@sharunkumar Thanks, I did a change that should preserve the old port!

PR #1392, opened 2024-04-05
Setup Playwright for end-to-end testing

aashu16:
I emailed a few months ago because I wanted to add e2e testing to the app. It has been a while (sorry) but here is my first PR.

What's been done?

- Set up Playwright.
- Added a few basic tests and some hard-coded data for mocking APIs (plan to remove this in the future, more below).

What's next?

- Add more tests: Adding tests for each view will take a while. If there are any bits of the app that you would like to prioritise, please let me know.
- Create abstractions: It's too early to do this, but once a few pages have achieved satisfactory coverage, would be ideal to spend some time on this.
- Use a live test server instead of mocking the Lemmy API: Definitely worth doing, but mocking does the job for now.

I intend to contribute regularly going forward. Depending on the number of testing scenarios and their complexity, I might split tests for a page/view into more than one PR.


vite.config.ts
```diff
@@ -71,6 +71,7 @@ export default defineConfig({
     BUILD_FOSS_ONLY: !!process.env.BUILD_FOSS_ONLY,
   },
   test: {
+    exclude: ["e2e/", "node_modules/"],
```
aeharding, line 74: This should fix `pnpm test`:
```suggestion
    exclude: ["**/e2e/**", "**/node_modules/**"],
```

aeharding (commented):
Hi! Thanks so much, this is a pleasant surprise! The PR looks great, just a small thing to get the build passing.

I'd love to get this running in CI - could you add a Github Action to `.github` folder? Otherwise, I am happy to look into it.

Also we should probably document in `README.md` but if you want to add in a follow up PR I am fine with that too.

aashu16:
> Hi! Thanks so much, this is a pleasant surprise! 

Appreciate the welcome. Looking forward to making meaningful contributions to the project.

> The PR looks great, just a small thing to get the build passing.

Can't believe I missed that. Fixed.

> I'd love to get this running in CI - could you add a Github Action to `.github` folder? Otherwise, I am happy to look into it.

Done.

> Also we should probably document in `README.md` but if you want to add in a follow up PR I am fine with that too.

Wondering if it might be better to wait until a somewhat acceptable test coverage has been achieved? Or did I completely misunderstand what you were suggesting?

aashu16:
> I mostly am asking because I wasn't sure how best to run e2e tests locally - but I found npx playwright test works well.

I changed Playwright's config to run off `vite preview` because it's significantly quicker - which is important for CI. For local testing, it's probably better to use the dev server and use the `--ui` flag. I will create a PR for this.

On a somewhat related note, do you have any preference regarding the size of PRs? If you usually squash and merge, I am guessing smaller and more frequent PRs would make sense.

aeharding:
> On a somewhat related note, do you have any preference regarding the size of PRs? If you usually squash and merge, I am guessing smaller and more frequent PRs would make sense.

The main reason I squash is to remove invalid/intermediate commits like "progress" or whatever that would add noise to main. Ideally every commit in main is a standalone bit of work. If you would prefer to merge without squashing for larger PRs I can totally do that, assuming the commits for the PR are cleaned up before merging (`git rebase -i HEAD~3` or whatever to remove the "progress"/"fix build"/"pr feedback" intermediate commits) so that every commit in the PR is a specific feature/bugfix/improvement.

aashu16:
That makes sense. I will try and keep PRs small and focused so we don't have to worry preserving commit history.

PR #1397, opened 2024-04-06
Docker: copy only the files required for build

sharunkumar:
- [x] build and run
- [x] check PWA install

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,8 @@ COPY patches ./patches
 RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile --ignore-scripts
 
 # Copy all source files
-COPY . ./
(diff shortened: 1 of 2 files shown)


sharunkumar:
Alternatively, I was wondering if we should just copy only the files needed for building, in 

https://github.com/aeharding/voyager/blob/main/Dockerfile#L29 ?

Dockerfile
```diff
@@ -26,7 +26,8 @@ COPY patches ./patches
 RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile --ignore-scripts
 
 # Copy all source files
-COPY . ./
+COPY build.sh disable_in_app_purchases.sh index.html vite.config.ts manifest.json tsconfig.json tsconfig.node.json ./
```
sharunkumar, line 29: build + running looks good when tested locally

aeharding (approved):
Cool, thanks!

Issue #1046, opened 2023-12-10
Modlog should show moderation reason


0xCmdrKeen:
**Is your feature request related to a problem? Please describe.**
As mentioned by @s38b35M5 in #995, the modlog currently only shows WHAT was moderated, but not WHY. The Lemmy API *does* provide a `reason` for each moderation item (unless the moderator didn't or couldn't add one, see #1045).

Additionally, many moderation items also leave out other important information (see [comment below](https://github.com/aeharding/voyager/issues/1046#issuecomment-1849228927)).

**Describe the solution you'd like**
Modlog entries should display the reason for each item, if available. Also, other relevant information, such as the community a post was deleted from, should be shown as well in order to make the modlog more useful and easier to understand.

**Describe alternatives you've considered**
Just don't show the moderation reason? IDK seems silly not to. It would help users to know what rule(s) they've broken so they can avoid further actions in the future.

**Additional context**
Voyager already seems to extract the `reason` from the API (see `src/features/moderation/logs/types/shared.ts:9`), but for some reason does not make use of it in `ModLogItem`. This shouldn't be very difficult from a technical point of view.

sallyFoster:
@0xCmdrKeen For some reason, for me, the community where the post was deleted from and the mod who deleted the post are not showing. The only things that are showing are the post title and the action taken. 

![image](https://github.com/aeharding/voyager/assets/104597302/57c63455-152d-4dae-9104-52cc0153ecb9)

0xCmdrKeen:
@aeharding happy to do some work on this one next week, unless you already have a plan for how you want to attack this.

I don't really have a good idea of how it might look like just yet, so probably just gonna play with a few variations until I find something that seems to click.

aeharding:
I've assigned it to you! I don't have many thoughts, other than I want to keep the general look and feel of it the same.

sallyFoster:
@0xCmdrKeen Did you get a chance to work on these?

0xCmdrKeen:
@sallyFoster yeah I played around some but haven't really had any brilliant ideas other than simply sticking the reason at the bottom like this:

![Screenshot](https://github.com/aeharding/voyager/assets/98132670/accdc106-0fe5-4f40-b68f-8fd557f610e9)

sallyFoster:
@0xCmdrKeen That looks fine to me.

sallyFoster:
@0xCmdrKeen  Is it possible to say where the user was banned from?

[...]

PR #1402, opened 2024-04-12
Revamped modlog with moderation reason

0xCmdrKeen:
By request, I rebased my old branch for #1046 to hopefully work with the recent code changes.

Had to restore a couple of functions that were removed or changed (`helpers/array.ts#notEmpty()` and `useIsAdmin()` with optional argument). Only tested briefly but seems to work fine.

### DIFF
--- a/src/features/moderation/logs/ModlogItem.tsx
+++ b/src/features/moderation/logs/ModlogItem.tsx
@@ -14,7 +14,7 @@ import removeComment from "./types/removeComment";
 import removeCommunity from "./types/removeCommunity";
 import removePost from "./types/removePost";
 import transferCommunity from "./types/transferCommunity";
-import { IonItem } from "@ionic/react";
+import { IonIcon, IonItem } from "@ionic/react";
 import { maxWidthCss } from "../../shared/AppContent";
 import Ago from "../../labels/Ago";
 import { useBuildGeneralBrowseLink } from "../../../helpers/routes";
@@ -37,17 +37,51 @@ import {
 } from "lemmy-js-client";
 import { styled } from "@linaria/react";
 
-const Contents = styled.div`
-  font-size: 0.875em;
-  width: 100%;
-
+const Container = styled.div`
   display: flex;
-  flex-direction: column;
-  gap: 6px;
-  padding: 8px 0;
+  gap: 1rem;
+
+  ${maxWidthCss}
+
+  padding: 0.5rem 0;
+
+  font-size: 0.875em;
+
+  strong {
+    font-weight: 500;
+  }
 `;
 
-const TitleLine = styled.div`
+const StartContent = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: 1rem;
+`;
+
+const Content = styled.div`
+  flex: 1;
+`;
+
+const Header = styled.div`
+  display: flex;
+  justify-content: space-between;
+
+  aside {
+    color: var(--ion-color-medium);
+    margin-left: auto;
+
+    display: flex;
+    align-items: center;
+    gap: 0.5rem;
+  }
+`;
+
+const Body = styled.div`
+  color: var(--ion-color-medium);
+  padding: 0.5rem 0;
+`;
+
+const Footer = styled.div`
   display: flex;
   justify-content: space-between;
 
@@ -57,9 +91,6 @@ const TitleLine = styled.div`
 `;
 
 const Title = styled.div``;
-const Message = styled.div`
-  color: var(--ion-color-medium);
-`;
 const Reason = styled.div``;
 
 interface ModLogItemProps {
@@ -67,10 +98,12 @@ interface ModLogItemProps {
 }
 
 export interface LogEntryData {
+  icon: string;
   title: string;
   when: string;
   message?: string;
   reason?: string;
+  expires?: string;
   by?: string;
   link?: string;
 }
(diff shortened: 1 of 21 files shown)


src/helpers/array.ts
```diff
@@ -1,5 +1,11 @@
 import { pullAt } from "lodash";
 
+export function notEmpty<TValue>(
```
aeharding, line 3: You can use lodash compact on the array instead.

src/features/moderation/logs/ModlogItem.tsx
```diff
@@ -14,80 +14,154 @@ import removeComment from "./types/removeComment";
...
   by?: string;
+  role?: ModeratorRole;
   link?: string;
 }
 
 function renderModlogData(item: ModlogItemType): LogEntryData {
   switch (true) {
     case "mod_remove_comment" in item:
-      return removeComment(item);
+      return removeComment(item as ModRemoveCommentView);
```
aeharding, line 136: Why were these `as` added? Typescript should be able to infer this

aeharding (commented):
Thanks for making this PR, I'll review it in detail soon!

src/helpers/array.ts
0xCmdrKeen: Oh gotcha. Okay, I'll fix it.

src/features/moderation/logs/ModlogItem.tsx
0xCmdrKeen: I made this months ago, it's possible that VS Code was giving me warnings, but perhaps that's changed. I'll review.

0xCmdrKeen: Yep, looks like it's no longer necessary. Removed.

src/features/moderation/useIsAdmin.tsx
```diff
@@ -1,6 +1,12 @@
+import { Person } from "lemmy-js-client";
 import { useAppSelector } from "../../store";
-import { isAdminSelector } from "../auth/siteSlice";
 
-export default function useIsAdmin() {
-  return useAppSelector(isAdminSelector);
+export default function useIsAdmin(person?: Person) {
```
aeharding, line 4: What is the purpose of this change?

src/helpers/useAppNavigation.ts
```diff
@@ -46,9 +47,13 @@ export default function useAppNavigation() {
   );
 
   const navigateToCommunity = useCallback(
-    (community: CommunityView) => {
+    (community: CommunityView | Community) => {
       pushRouteIfNeeded(
-        buildGeneralBrowseLink(`/c/${getHandle(community.community)}`),
+        buildGeneralBrowseLink(
+          `/c/${getHandle(
```
aeharding, line 53: Can you use `buildCommunityLink` from `appLinkBuilder`? Like `buildGeneralBrowseLink(buildCommunityLink(community))`.

(Sorry I know it's not done everywhere, I'm slowly switching over to it.)

You can make `buildCommunityLink` take a `CommunityView` if it would help!

src/features/moderation/logs/ModlogItem.tsx
```diff
@@ -99,24 +156,67 @@ function renderModlogData(item: ModlogItemType): LogEntryData {
...
-          <aside>
-            {by && <span>{by} · </span>}
-            <Ago date={when} />
-          </aside>
-        </TitleLine>
-        <Message>{message}</Message>
-      </Contents>
+      <Container>
+        <StartContent>
+          <IonIcon icon={icon} color="medium" />
```
aeharding, line 195: I see you're using the styles from InboxItem, which is great! Do you think its worth abstracting them into a component that takes `icon`? Perhaps for another task?

src/features/moderation/useIsAdmin.tsx
0xCmdrKeen: I needed a way to determine admin status for arbitrary users so I can properly highlight them in the modlog (see ModLogItem.tsx:171). 

This is basically the previous version of this function (from before it was replaced with a static selector in ead60d3), augmented by an optional `person` parameter.

Since it's not used anywhere else right now, I can make this a private function in `ModLogItem` instead and restore this helper to its last state if you prefer.

src/helpers/useAppNavigation.ts
0xCmdrKeen: Sure, no problem.

src/features/moderation/logs/ModlogItem.tsx
0xCmdrKeen: Good question. I did indeed copy some of the styled components from there in order to have the design stay consistent across the app, thinking it would make sense to have the modlog look similar to a view that users will already be familiar with, that way it would be easier to read.

The reason I didn't extract them into shared components is because they are still separate features and I didn't want to unecessarily entangle them in case they evolve somewhat differently in the future.

src/features/moderation/useIsAdmin.tsx
aeharding: I see. How about something like `useAppSelector(!person ? isAdminSelector : state => state.site.response?.admins?.some(
      (admin) => admin.person.id === person?.id
    ))` instead?

0xCmdrKeen: Sounds good, can do.

aeharding (approved):
Looks good, just failing CI

aeharding:
DOH, I don't know why Github didn't wait for successful CI before merging :(

PR #1408, opened 2024-04-13
Display Error Message in Toasts

### LINKED ISSUE #1399: Show the right error message to the user
Sometimes a random error occurs and the user has no way to find out what went wrong unless they dig into the console / network logs:

image

Error in browser console:

image

Seems like in this case, its an invalid post title

I can work on a PR for this

sharunkumar:
Fixes #1399


src/helpers/error.ts
```diff
@@ -0,0 +1,7 @@
+export function presentErrorMessage(message: string, error: unknown): string {
```
aeharding, line 1: Hey thanks for the PR!

I see your PR exposes the raw error from lemmy. Voyager's current error handling approach is to handle each error and convert to english. This also adds the ability to compose specific, helpful error messages for more casual Voyager/lemmy users. You can see that in `helpers/lemmyErrors.ts`

I think we could make a function, `getCreatePostErrorMessage` similar to the existing `getVoteErrorMessage`.

We could also do that for other errors, as I see in your PR.

sharunkumar: I like the implementation of `getVoteErrorMessage` but I would make a change where I would still like to add the raw error message in the fallback condition. This way, even if there is an error which we have not handled (mapped to a user-friendly message) there would at least be some information available to the user

sharunkumar: Let me know if the changes in https://github.com/aeharding/voyager/pull/1408/commits/86fb64091f78a77f487d73ceee2e7aee4495ef57 works. Motivation here is to provide some message to the user rather than a generic one

PR #1473, opened 2024-05-11
Use `thumbnail_url` as fallback

### LINKED ISSUE #1472: This specific image post doesn't load
Post in question: https://vger.app/profile/lemmy.ml/c/196@lemmy.blahaj.zone/comments/15488079

image

### Observations

- The post loads in web view: https://pawb.social/post/9689361
  - Thumbnail link: https://pawb.social/pictrs/image/2060d98d-fb9e-47bf-9074-cdf0f8a77199.jpeg?format=webp&thumbnail=256
  - Image Link: https://pawb.social/pictrs/image/2060d98d-fb9e-47bf-9074-cdf0f8a77199.jpeg
- The `post` data that we received is having the `thumbnail_url` as https://lemmy.blahaj.zone/pictrs/image/cc19d66b-2837-4f33-9480-ceac04bc7f8b.jpeg, which seems to be the cause of failure (link seems to be a 400 - bad request)

In `src/features/media/gallery/Media.tsx#getPostMedia`, the `if (post.post.thumbnail_url) return post.post.thumbnail_url` seems to be repeated

image

I guess the idea was to load the thumbnail if its available or fall back to the post url. but in this specific case, the `thumbnail_url` seems to be invalid (pawb.social seems to be using a different link for displaying the thumbnail)

sharunkumar:
Possibly fixes #1472

### DIFF
--- a/src/features/media/gallery/Media.tsx
+++ b/src/features/media/gallery/Media.tsx
@@ -39,8 +39,6 @@ const Media = forwardRef<
 export default memo(Media);
 
 export function getPostMedia(post: PostView): string | undefined {
-  if (post.post.thumbnail_url) return post.post.thumbnail_url;
-
   if (post.post.url && isUrlMedia(post.post.url)) {
     return post.post.url;
   }


aeharding:
Hi!

This is specifically loaded first for privacy reasons because the thumbnail url is usually cached on your own instance, so the client doesn't make requests to arbitrary hosts.

So we should probably look for another approach.

sharunkumar:
Closing in favor of https://github.com/aeharding/voyager/pull/1477

Issue #1505, opened 2024-05-30
Small fix for Redgifs embedded media


net00-1:
**Is your feature request related to a problem? Please describe.**
- Many Redgifs uploads are done using the `v3.redgifs.com` subdomain.  Voyager treats these links as regular link posts instead of embedding the media. For example https://lemmynsfw.com/post/12153533 (NSFW)

**Describe the solution you'd like**
- Redgifs link detection seems to be done in `src/features/media/external/redgifs/helpers.ts` with regex:
  - `^https:\/\/(?:www\.)?redgifs.com\/watch\/([a-z]+)`

- I added v3 subdomain like so `https:\/\/(?:www\.|v3\.)?redgifs.com\/watch\/([a-z]+)` within my fork. After building the app for iOS I see that links with v3 subdomain are now properly embedded as well. I tested this with iPad and iPhone on iOS 17.5.1

**Describe alternatives you've considered**
- N/A

**Additional context**
- My fork https://github.com/net00-1/voyager (built in codemagic.io, please ignore `codemagic.yaml`)

PR #1506, opened 2024-05-30
Small Redgifs regex fix

net00-1:
Adds `v3` to Redgifs video filtering regex, which allows media to be embedded instead of displayed as regular link.


net00-1:
For issue #1505

aeharding (approved):

aeharding:
Thanks!

PR #1515, opened 2024-06-06
Use device mode specific share icon

### LINKED ISSUE #1325: Use Android share icon
Android has a different share icon than iOS. It would be cool if we could switch icons dynamically depending on the device theme.

aeharding:
Resolves #1325


sharunkumar:
This was one of the things I was patching on my own version :laughing: 
Thanks for implementing this!

sharunkumar:
Actually, I was having `shareSocialOutline` in the iOS theme itself

Is it okay if I raise a PR which would let the user configure/override the theme specific icon?

aeharding:
I think that's a bit too fine grained for a setting. Probably best to keep that in a fork for people that prefer that level of customization. :)

PR #1577, opened 2024-08-10
Improve support for testing (http, custom ports, env variables)

### LINKED ISSUE #1576: Make it easier to run Playwright tests
**Is your feature request related to a problem? Please describe.**

I am trying to update the e2e GH action to set up a local Lemmy stack in order to allow Playwright tests to make real API calls instead of mocking them.

To make this work,
- The local Lemmy API server needs TLS configuration (using a reverse proxy and self-signed certificates) since the url builder always prepends `https://` to server names provided.
- Custom Lemmy servers can only be provided by configuring the `/_config` endpoint.

These problems are not insurmountable and I have managed to come up with a solution that works but not without adding unnecessary complexity to the testing stack without improving the quality of the tests. In the interest of simplifying the testing environment, I would like to suggest a couple of changes.

These changes could possibly allow users to misconfigure Voyager instances. It should, however, be easy enough to ensure that one would have to go out of their way to be able to do so.

**Describe the solution you'd like**

The following changes should be restricted to development and CI environments.
1. An easier way to provide Voyager with a list of Lemmy servers - either hardcoded or via an env var.
2. Base URL builder should be able to use `http://`.

Happy to discuss alternatives and/or concerns.

aeharding:
Resolves #1576

Now you can set the following (testing purposes only!):

- `VITE_FORCE_LEMMY_INSECURE=1` to force all lemmy instance connections to http instead of https.
- `VITE_CUSTOM_LEMMY_SERVERS=localhost:3001,localhost:3002` a comma separated list of default lemmy servers. If _config is found it will be preferred.

### DIFF
--- a/src/routes/common/ActorRedirect.tsx
+++ b/src/routes/common/ActorRedirect.tsx
@@ -33,7 +33,7 @@ function ActorRedirectEnabled({ children }: ActorRedirectProps) {
   const [first, second, _wrongActor, ...urlEnd] = location.pathname.split("/");
 
   // no need to redirect if url doesn't have actor
-  if (!_wrongActor || !_wrongActor.includes(".")) return page;
+  if (!_wrongActor || !isPotentialActor(_wrongActor)) return page;
 
   return (
     <Redirect
@@ -41,4 +41,13 @@ function ActorRedirectEnabled({ children }: ActorRedirectProps) {
       push={false}
     />
   );
+}
+
+function isPotentialActor(host: string) {
+  if (host.includes(".")) return true;
+  if (host.includes(":")) return true;
+
+  if (host.startsWith("localhost")) return true;
+
+  return false;
 }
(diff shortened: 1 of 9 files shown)


src/routes/common/ActorRedirect.tsx
```diff
@@ -42,3 +42,12 @@ function ActorRedirectEnabled({ children }: ActorRedirectProps) {
     />
   );
 }
+
+function isPotentialActor(host: string) {
```
aeharding, line 46: @aashu16 see this function - we can add additional hostnames here if needed.

Right now, should support:

- Any hostname with dot (domain) or colon (port number)
- localhost

aashu16: Not sure if there is any point in being stricter with these checks.

Marking as resolved because this works for now.

aashu16 (approved):
LGTM.

Tested this locally and works great. Thanks for pushing this out so quickly.

aeharding:
Thanks for reviewing!

PR #1579, opened 2024-08-11
Add .editorconfig

aashu16:
Adding an [EditorConfig](https://editorconfig.org/) which is in line with the repo's prettier config and works out-of-the-box on most editors.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+root = true
+
+[*.{js,ts,yaml,json}]
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,6 @@
+root = true
+
+[*.{js,ts,yaml,json}]
```
aeharding, line 3: Do you also need to specify jsx, tsx?

.editorconfig
```diff
@@ -0,0 +1,6 @@
...
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
```
aeharding, line 6: Could probably add

`insert_final_newline = true`

too!

.editorconfig
aashu16: Added.

Was trying to limit the impact on other contributors, but this shouldn't really break anybody's workflow.

aeharding (approved):

PR #1585, opened 2024-08-12
Update e2e action to use a better testing setup

aashu16:
Some things I want to highlight or would like opinions/thoughts on.
- The e2e action, as of this commit, will not run on `push` and `pull_request` triggers. It will have to be manually triggered.
- The idea is to ensure that the tests are somewhat stable before adding back these removed triggers. In the meantime, there will be failed runs of the action but they will become increasingly infrequent with time.
- PRs will be fairly frequent for the next few weeks and tests might undergo several changes (and refactors) before they become stable. This will probably involve breaking from this repo's convention of squash merging PRs but I will try to avoid any superfluous commits.
- If you feel there are user stories that should take priority, please let me know.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,17 +1,9 @@
 name: Run e2e tests
-on:
-  push:
-    branches:
-      - main
-  pull_request:
-    branches:
-      - main
+on: workflow_dispatch
 jobs:
   playwright:
     name: Run Playwright tests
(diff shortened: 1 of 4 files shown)


aeharding (approved):
This is exciting to test against lemmy api. thanks!

aeharding:
Feel free to merge

aashu16:
One of the checks is stuck on pending, will need bypassing. Can you merge this, @aeharding?

PR #1593, opened 2024-08-22
Override lemmy server URL when 'VITE__TEST_MODE' is set

aashu16:
After the user logs in, `state.auth.currentInstance` changes from `<hostname>:<PORT>` to `<hostname>` [1]. Setting `VITE__TEST_MODE=1` will override this behaviour with a couple of small changes and no side-effects. Using this var name because similar overrides might be needed elsewhere in the future and this should cover most (or all) of them.

Also updated the build step in e2e action.

[1]: [Lemmy sets the `iss` claim to hostname](https://github.com/LemmyNet/lemmy/blob/63a686d390e441216718a3d64408336409890791/crates/api_common/src/claims.rs#L48) when creating JWTs. `authSlice.addJwt` updates `state.auth.currentInstance` with the value of this claim.

### DIFF
--- a/src/features/auth/authSelectors.ts
+++ b/src/features/auth/authSelectors.ts
@@ -49,8 +49,11 @@ export const instanceSelector = createSelector([handleSelector], (handle) => {
   return getInstanceFromHandle(handle);
 });
 
-export const urlSelector = (state: RootState) =>
-  instanceSelector(state) ?? state.auth.connectedInstance;
+export const urlSelector = (state: RootState) => {
+  if (import.meta.env.VITE__TEST_MODE) return state.auth.connectedInstance;
+
+  return instanceSelector(state) ?? state.auth.connectedInstance;
+};
 
 export const clientSelector = createSelector(
   [urlSelector, jwtSelector],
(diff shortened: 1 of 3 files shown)


aashu16:
Missed an import. Added in the second commit.

src/features/auth/authSelectors.ts
```diff
@@ -49,8 +49,11 @@ export const instanceSelector = createSelector([handleSelector], (handle) => {
   return getInstanceFromHandle(handle);
 });
 
-export const urlSelector = (state: RootState) =>
-  instanceSelector(state) ?? state.auth.connectedInstance;
+export const urlSelector = (state: RootState) => {
+  if (import.meta.env.VITE__TEST_MODE) return state.auth.connectedInstance;
```
sharunkumar, line 53: curious: shouldn't this be implemented by mocking when running the tests? :thinking:

aeharding: That would be a nice improvement

aeharding (approved):
Sorry for the delay! Need to get through all these PRs :)

src/features/auth/authSelectors.ts
aashu16: Sorry for taking this long to respond.

Not entirely clear on what we would want to mock here, @sharunkumar. Could you please expand on it? Feel free to chime in, @aeharding.

PR #1594, opened 2024-08-24
Fix: Remove block button when long-pressing your own handle

### LINKED ISSUE #1592: Option to block yourself appears
**Describe the bug**
When using the client for Android, I can hold down my own username, and it still prompts me to `Block User`. I'm too scared to try this to see what this actually does, but I do not believe this is a feature of Lemmy.

**To Reproduce**
Navigate to your username e.g. in a comment or in a post, and hold down your username. 

**Expected behavior**
You probably should not be given the option to block your own account.

**Smartphone:**
- Device: Galaxy S23 Ultra
- OS: Android 14
- Version 2.17.0

sharunkumar:
Fixes #1592


arunsathiya (approved):

src/features/labels/links/PersonLink.tsx
```diff
@@ -56,47 +59,48 @@ export default function PersonLink({
...
-                });
-                throw error;
-              }
-
-              presentToast(buildBlocked(!isBlocked, getHandle(person)));
-            })();
+    const isCurrentUser = currentUserHandle === getRemoteHandle(person);
+    const buttons: Button[] = [
+      isCurrentUser
+        ? "That's me!"
```
aeharding, line 76: Instead of this, can we hide the button? I realize it will just have the cancel button. However in the future if options are added this may be OK.

The alternative is to disable long press completely, or provide an alert saying you can't block yourself when you click block user.

sharunkumar: I'll go with removing the button for now

src/features/labels/links/PersonLink.tsx
```diff
@@ -56,47 +60,47 @@ export default function PersonLink({
     (admin) => admin.person.actor_id === person.actor_id,
   );
   const { hideUsernames } = useContext(ShareImageContext);
+  const currentUserHandle = useAppSelector(userHandleSelector);
```
aeharding, line 63: Because this isn't used for render, can you define this within the `onCommunityLinkLongPress` function?

```
const currentUserHandle = userHandleSelector(state);
```

aeharding (approved):
Thanks!

Issue #1588, opened 2024-08-15
Search: all instance posts


jujubeja:
**Is your feature request related to a problem? Please describe.**
As user
I want to add multiple instances of lemmy
So that you can view posts from all instances that do not follow each other

**Describe the solution you'd like**

Enable full instance search
https://lemmy.world/search?q=%s
https://lemm.ee/search?q=%s

Load List: import instance list by url .txt and .toml and .json
```
https://raw.githubusercontent.com/aeharding/voyager/main/instances.toml
https://github.com/aeharding/voyager/blob/main/instances.toml
```

**Additional context**
full instance list
https://data.lemmyverse.net/
https://lemmy.fediverse.observer/list

aeharding:
Hi there!

As mentioned on #1600 unfortunately Play Store limits the number of domains on the list, so I can't add everything.

PR #1600, opened 2024-09-10
Add: source from data.lemmyverse.net All instances

atlantis451:
Add source from data.lemmyverse.net All instances
#1588


aeharding:
Hi, thanks for the PR!

The reason there is only a few, is because the Play Store restricts the number of instances we can have on this list.

So unfortunately I have to close this PR. Sorry

PR #1607, opened 2024-09-27
Add dark & tinted icon variants for default app icon

fer0n:
Most dark versions aren't really all that dark, but I couldn't come up with anything non-white that I found still looked acceptable, but perhaps someone else has success there.

Also, here's the updated source file: 
[Voyager icon.zip](https://github.com/user-attachments/files/17187039/Voyager.icon.zip)

What do you think? (tined left, dark mid & right variants) 
I sometimes went with the opaque and sometimes the transparent background, for some I think that makes more sense than for others. Do you have any preference there?
Picsew_20240928195257


aeharding:
Looks great 😍 I think I like the dark mode grey BG better, it seems to match better with other iOS icons.

I'll leave it to your discretion, though. We can also tweak in the future. So let me know when you're happy and I'll get this in.

fer0n:
I've been staring at them way too long. I think isolated the ones with a colorful background look nicer, but the gray ones fit in better. Psych fits in best, holidays is also not too bad.

For the other ones I'm still struggling, but it's probably best to make them fit in more. For planetary I'll stick with the custom background, it has too much color everywhere and I don't think it matches the gray very well.

As you said, we can always tweak it later. As far as I'm concerned you can merge it, but if you want any changes I'm fine with that as well.

aeharding (approved):

mrtzlbm:
Love these! Are they supposed to work already with version >2.18? I tried both the app store as well as the testflight version and in neither did the dark or tinted icons get applied (on iOS 18.0.1).

fer0n:
@mrtzlbm I don't think they're in a released version yet

mrtzlbm:
Hi @aeharding, are these new app icon styles supposed to work already with Voyager version >2.18 since this has been merged and mentioned in the release notes? I tried both the App Store as well as the TestFlight version and in neither did the dark or tinted icons get applied (iOS 18.0.1). Thanks!

aeharding:
Yes, they are. Can you send a screenshot of what you're seeing? Also, try restarting your device, sometimes iOS caches icons strangely

fer0n:
Restart definitely doesn't hurt, but iOS 18 seems to have fixed that issue, before it never changed without it. 

The icon itself changes for me, but the dark version doesn't show up. Does it work for you @aeharding?

aeharding:
Works fine.
IMG_D719A3D2F4D9-1

fer0n:
Is that the App Store Version? I tried both TestFlight and App Store, restarted and changed icons, but the dark version never shows up.

aeharding:
Works in dev, not in store/testflight. 🤦‍♂️ I'm not sure what the issue is.

aeharding:
I figured it out, the build agent was using an old version of Xcode!

mrtzlbm:
@aeharding Awesome you found the bug already! Here is my screen recording anyway: https://imgur.com/a/gezNXHY 😅

fer0n:
It works now, thanks :)

aeharding:
Really love these. Especially the Pride flag, it has a nice pop. Thanks again @fer0n

Issue #1652, opened 2024-11-04
While crossposting, communities are truncated/missing compared to lemmy-ui


kminardo:
**Describe the bug**
When attempting to crosspost to other communities, only the top 10 communities (by subscriber count) are available for selection. This is a problem as other communities may be further down the results and not available to users of the app.

Specifically, this is a problem for me, as I cannot crosspost to a local instance community that is outside of the top 10 without using lemmy-ui.

**To Reproduce**
1. Go to a post
2. Open the crosspost modal
3. Select 'Community'
4. Search for "Linux" or "Technology"
5. Only top 10 community results are visible/available

**Expected behavior**
I expect to be able to select from the same 20 communities available using lemmy-ui

**Screenshots**

Full Voyager List:
<img src="https://github.com/user-attachments/assets/293e0cb7-30bc-4b26-b563-49824fbd65d6" width="250">
<img src="https://github.com/user-attachments/assets/9385c457-6254-47a5-8b88-0ce380d9b11e" width="250">

Full Lemmy-UI List:
<img src="https://github.com/user-attachments/assets/83bc2579-f4b4-4e3e-ad69-2853a3ce5c7b" width="450">

**Desktop (please complete the following information):**
**Smartphone (please complete the following information):**
Effects all

**Additional context**
Some investigation pointed me towards this query:
https://github.com/aeharding/voyager/blob/d900c3b3d3fb7aaf334bd97c6e1f547a44fa6fa2/src/features/shared/selectorModals/CommunitySelectorModal.tsx#L18

There is no `limit` property being passed, resulting in the API returning 10 results. 

When I use Lemmy-UI, this API url is built with limit=20, returning more communities:
`https://sh.itjust.works/api/v3/search?q=linux&type_=Communities&sort=TopAll&listing_type=All&page=1&limit=20`

PR #1653, opened 2024-11-04
Update CommunitySelectorModal API query to use global limit

kminardo:
Change to correct behavior documented @ https://github.com/aeharding/voyager/issues/1652

I chose to use the existing LIMIT constant (50) instead of 20 - but can correct this to exactly match lemmy-ui, if that's preferred.


aeharding (approved):
Thanks, looks great! Apologies if it take a while to release, the app is currently in a state of flux :)

PR #1704, opened 2024-11-12
fix(docker): pre-build script

### LINKED ISSUE #1703: Docker: pre-build script does not actually run in docker build
https://github.com/aeharding/voyager/actions/runs/11791484474/job/32843572978#step:7:607

image

possibly `bash` is not available in the base image

suggested fix: change it to `sh "$SETUP_SCRIPT"`

also the path needs to be updated: `SETUP_SCRIPT="./scripts/disable_in_app_purchases.sh"`

I've checked it locally, will raise a PR in a bit

sharunkumar:
Fixes #1703

verified that it works in CI: https://github.com/aeharding/voyager/actions/runs/11805078576/job/32886733829?pr=1704#step:7:706

### DIFF
--- a/scripts/prebuild.sh
+++ b/scripts/prebuild.sh
@@ -1,11 +1,11 @@
 #!/bin/sh
 
-SETUP_SCRIPT="./disable_in_app_purchases.sh"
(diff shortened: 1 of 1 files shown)


aeharding:
Thanks!

PR #1724, opened 2024-11-15
chore: upgrade dependencies

aeharding:


sharunkumar:
@aeharding have you considered using renovate for automating dependency upgrades? https://docs.renovatebot.com/

aeharding:
I've tried it but in my experience it's easier/safer/quicker to update myself.

PR #1860, opened 2025-02-21
Feat: Collapse downvoted comments by default when setting (default off) is set

### LINKED ISSUE #1838: Auto hide downvoted comments
I couldn't find it in settings, but it would be nice if downvoted comments could be hidden or auto collapsed. (Based on n number of downvotes perhaps).

It would mean lower visibility for trolls which would lower their engs ment / reward they get out of it.

jameslounds:
Closes #1838 

This is my first contribution, so it's very possible I've misunderstood something or not followed conventions in the project. As well as more substantive feedback, nitpicks are more than welcome - they help me learn!

<details><summary>About the `ppm-lcck.yaml` update</summary>
<p>

Also not entirely sure why, but the `pnpm-lock.yaml` needed updating. `pnpm i --frozen-lockfile` gave an error 
```
❯ pnpm i --frozen-lockfile
 ERR_PNPM_LOCKFILE_CONFIG_MISMATCH  Cannot proceed with the frozen installation. The current "patchedDependencies" configuration doesn't match the value found in the lockfile

Update your lockfile using "pnpm install --no-frozen-lockfile"
```
I I've put the changes to `pnpm-lock.yaml`  into a separate commit I can revert on request .

</p>
</details> 

The setting is off by default.

I've added `hasUserCollapsed` which is true if  the value for `commentId` in `state.comment.commentCollapsedById` has been defined. If it hasn't we're in the initial state of the comment (show unless downvoted & setting is set). If it has, the user must have shown/hidden the comment so we should use the state from `state.comment.commentCollapsedById` rather than the default.

### DIFF
--- a/src/features/comment/inTree/CommentTree.tsx
+++ b/src/features/comment/inTree/CommentTree.tsx
@@ -32,14 +32,27 @@ export default function CommentTree({
   baseDepth,
 }: CommentTreeProps) {
   const dispatch = useAppDispatch();
+  const commentId = comment.comment_view.comment.id;
+  const hasUserCollapsed = useAppSelector((state) =>
+    Object.hasOwn(state.comment.commentCollapsedById, commentId),
+  );
   const collapsed = useAppSelector(
-    (state) =>
-      state.comment.commentCollapsedById[comment.comment_view.comment.id],
+    (state) => state.comment.commentCollapsedById[commentId],
   );
   const tapToCollapse = useAppSelector(
     (state) => state.settings.general.comments.tapToCollapse,
   );
   const { activePageRef } = useContext(AppContext);
+
+  const downvotedCollapsedByDefault = useAppSelector(
+    (state) => state.settings.general.comments.downvotedCollapsedByDefault,
+  );
+
+  const isDownvoted =
+    comment.comment_view.counts.upvotes < comment.comment_view.counts.downvotes;
+  const shouldBeCollapsed = hasUserCollapsed
+    ? collapsed
+    : downvotedCollapsedByDefault && isDownvoted;
 
   // Comment context chains don't show missing for parents
   const showMissing = (() => {
@@ -68,7 +81,7 @@ export default function CommentTree({
         depth={comment.absoluteDepth - baseDepth}
         absoluteDepth={comment.absoluteDepth}
         key={comment.comment_view.comment.id}
-        collapsed={collapsed || fullyCollapsed}
+        collapsed={shouldBeCollapsed || fullyCollapsed}
         comment={comment}
       />
     );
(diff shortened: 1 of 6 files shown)


src/features/comment/inTree/CommentTree.tsx
```diff
@@ -32,15 +32,28 @@ export default function CommentTree({
   baseDepth,
 }: CommentTreeProps) {
   const dispatch = useAppDispatch();
+  const commentId = comment.comment_view.comment.id;
+  const hasUserCollapsed = useAppSelector((state) =>
```
aeharding, line 36: Remove this selector and check if `collapsed == null`

src/features/comment/inTree/CommentTree.tsx
```diff
@@ -32,15 +32,28 @@ export default function CommentTree({
...
   const tapToCollapse = useAppSelector(
     (state) => state.settings.general.comments.tapToCollapse,
   );
   const { activePageRef } = useContext(AppContext);
 
+  const downvotedCollapsedByDefault = useAppSelector(
+    (state) => state.settings.general.comments.downvotedCollapsedByDefault,
+  );
+
+  const isDownvoted =
```
aeharding, line 51: Check if `counts.score < 0` instead

src/features/settings/general/comments/DownvotedCollapsedByDefault.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+  return (
+    <IonItem>
+      <IonToggle
+        checked={downvotedCollapsedByDefault}
+        onIonChange={(e) =>
+          dispatch(setDownvotedCollapsedByDefault(e.detail.checked))
+        }
+      >
+        Collapse Downvoted Comments
```
aeharding, line 21: I was initially confused by this! I thought it would collapse when I downvote. Instead, it collapses comments with negative cumulative vote (which makes more sense!).

I think we should come up with a different setting that avoids this confusion. Some options...

as a switch, "Auto Collapse Negative Votes"

or, as a dropdown with options:

"Collapse Downvoted Comments..."

- "When Negative"
- "When Highly Downvoted" (some cool algorithm we build hehe)
- "Never" (default)

I personally prefer the latter, and it opens the door for more customization. What are your thoughts?

src/features/settings/settingsSlice.tsx
```diff
@@ -228,6 +229,7 @@ const baseState: SettingsState = {
   general: {
     comments: {
       collapseCommentThreads: OCommentThreadCollapse.Never,
+      downvotedCollapsedByDefault: false,
```
aeharding, line 232: nit: remove 'ByDefault'

aeharding (changes requested):
Thanks for the PR! I tried it out. I really like it, seems to clean up the comments section a good amount.

However, I did find a bug: The first time I tap to expand an auto collapsed comment, it doesn't expand. It does on the second try, though.

LMK if you have any questions I'm happy to help

PR #1885, opened 2025-03-09
chore: move `eslint-plugin-perfectionist` to dev dependencies

melusc:
Like the other eslint dependencies, it is a dev dependency


aeharding (approved):
Thanks!

PR #2287, opened 2026-06-13
feat: decorate markdown syntax in select text modal

### LINKED ISSUE #2283: Select text should have same syntax formatting as rich text editor
title

aeharding:
closes #2283


aeharding:
nit: syntax highlighting only if rich text editor turned on

PR #2355, opened 2026-08-30
Fix blank virtual lists after tab switches

aeharding:


aeharding:
Superseded by upstream Virtua regression issue https://github.com/inokawa/virtua/issues/940 and its standalone reproduction. The dependency patch and timing-dependent Voyager test should not be merged.

